PR #1, opened 2016-10-06
functions: more reliable systemd detection logic

davide125:
On cgroup2 systems the current test for systemd doesn't work, as the cgroup hierarchy is setup differently. This results in the `systemctl_redirect` function not working properly, which breaks the redirection logic and leads to subtle bugs and general sadness. This PR changes the logic to just check for `/run/systemd/system`, which is independent from the cgroup hierarchy and is the recommended way to detect systemd by upstream.


lnykryn:
Looks good to me. Thanks!

PR #2, opened 2016-10-07
Provide a mechanism to prevent network-scripts from messing with sysctls

jaymzh:
network-scripts makes odd assumptions, like if I don't want SLAAC (i.e.
if I turn off AUTOCONF) that I also don't want RA. This is not true,
it's common to have a static local address, but to get my gateway
through RA.

This is one solution. I will also be sending a more specific solution to
enable the above directly, but having a way to tell network-scripts
to not mess with systctls seems generally useful.

### DIFF
--- a/sysconfig/network-scripts/ifdown-ipv6
+++ b/sysconfig/network-scripts/ifdown-ipv6
@@ -62,10 +62,12 @@ if [ $? != 0 -a $? != 11 ]; then
 	exit 1
 fi
 
-# Switch some sysctls to secure mode
-/sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.forwarding=0 >/dev/null 2>&1
-/sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.accept_ra=0 >/dev/null 2>&1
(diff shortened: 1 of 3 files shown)


sysconfig/network-scripts/ifdown-ipv6
```diff
@@ -62,10 +62,12 @@ if [ $? != 0 -a $? != 11 ]; then
 	exit 1
 fi
 
-# Switch some sysctls to secure mode
-/sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.forwarding=0 >/dev/null 2>&1
-/sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.accept_ra=0 >/dev/null 2>&1
-/sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.accept_redirects=0 >/dev/null 2>&1
+if [ ! "$IPV6_SET_SYCTL" = "no" ]; then
```
pavlix: Did you mean `IPV6_SET_SYSCTL` instead of `IPV6_SET_SYCTL`?

PR #3, opened 2016-10-07
[1/3] Remove tabs and fix spacing: ipv6 scripts

jaymzh:
The network-scripts are a mix of tabs and spaces (!!) as well as
completely inconsistent indentation (some 2, some 4).

This is a series of 3 patches to move the entire of network-scripts to
2-space indentation for consistency so I can develop on the codebase
without my brain bleeding. I don't feel strongly about 2 vs 4... 2 was
just most common in the first file I picked, so I used that.

This set is in 3 sets for easier reviewing and easier reverting if
necessary:

* 1/3 - ipv6 files
* 2/3 - network-functions - it needed it's own
* 3/3 - ifup/ifdown files

### DIFF
--- a/sysconfig/network-scripts/ifdown-ipv6
+++ b/sysconfig/network-scripts/ifdown-ipv6
@@ -58,80 +58,81 @@ ipv6_test testonly || exit 0
 # Test device status
 ipv6_test_device_status $DEVICE
 if [ $? != 0 -a $? != 11 ]; then
-	# device doesn't exist or other problem occurs
-	exit 1
+    # device doesn't exist or other problem occurs
+    exit 1
 fi
 
 if [ ! "$IPV6_SET_SYCTL" = "no" ]; then
-  # Switch some sysctls to secure mode
-  /sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.forwarding=0 >/dev/null 2>&1
-  /sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.accept_ra=0 >/dev/null 2>&1
-  /sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.accept_redirects=0 >/dev/null 2>&1
+    # Switch some sysctls to secure mode
+    /sbin/sysctl -e -w net.ipv6.conf.$SYSCTLDEVICE.forwarding=0 >/dev/null 2>&1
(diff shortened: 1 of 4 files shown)


jaymzh:
Note that this entire series and the non-whitespace patches will conflict with each other... so either you have to merge the whitespace ones and then I rebase the other two, or you merge the other two and then I rebase these 3... I don't really care which... just let me know.

Also, I need to figure out the best way to test these PRs. I'll probably do that next week...

lnykryn:
Well I am also sick with the current formatting, but to be honest I am lazy and this patch will make backports and git blame a little bit harder.

lnykryn:
Scratch that. As Michal pointed out, I just read the manpage for git blame and use -w.

lnykryn:
I don't have strong preference either, but can't we go for 8 spaces? I am really used to that from systemd :-).

jaymzh:
I... think you're joking?

systemd isn't written in shell... and 8 spaces is ... a LOT. Definitely not common practice... I could be convinced for 4, but 8 is excessive.

I'll re-base this on the other work you merged and test it this week.

lnykryn:
Ok, four should be fine ;-)

jaymzh:
rebased and moved to 4 spaces.

`git show -w` shows no meaningful changes.

jaymzh:
(don't merge these until I have a chance to test... there's weird places with regexes with magic numbers of spaces and stuff... )

lnykryn:
What I would really love to see is to run some diff-scans on pull requests with shellcheck. I was running them manually and it saved my ass couple of times.

lnykryn:
Hmm, it is really easy to setup travis with shellcheck and bashate, but by default the output is pretty huge. https://travis-ci.org/lnykryn/initscripts

jaymzh:
Yeah, I'm a big fan of shellcheck, I'll run it through there. The problem with putting it in Travis is it has a fair number of false-positives where it wants you do not quote or quote a thing that actually is wrong (will give you the wrong result).

jaymzh:
I'll test this on an actual system as well as run it through shellcheck (after I finish rebasing and fixing 3/3).

jaymzh:
OK I've tested this entire patch series with reboots and network restarts and ifup/ifdowns and I'm confident it's sane.

I did run shellcheck and found that it has lots of good advice, but I'd like to keep these particular PRs to just whitespace so that it's easy to track real changes separately. I will follow up with additional PRs for some shellcheck stuff if that's desired. There's one bug it caught I'll follow up with a PR immediately.

PR #4, opened 2016-10-07
[2/3] Remove tabs and fix spacing: network-functions

jaymzh:


jaymzh:
Rebased and updated to 4 spaces. `git show -w` shows only the changes where we removed useless trailing semi-colons, and where we unified where opening `{`s go on functions.

PR #9, opened 2016-10-11
ipv6_wait_tentative(): exit immediately when dadfailed occurs

deekej:
This pull request relates to RHBZ [#1378917](https://bugzilla.redhat.com/show_bug.cgi?id=1378917).

### DIFF
--- a/sysconfig/network-scripts/network-functions-ipv6
+++ b/sysconfig/network-scripts/network-functions-ipv6
@@ -1048,6 +1048,7 @@ ipv6_wait_tentative() {
     local fn="ipv6_wait_tentative"
     local device=$1
     local countdown=30
+    local ip_output=""
 
     if [ -z "$device" ]; then
         net_log $"Missing parameter 'device' (arg 1)" err $fn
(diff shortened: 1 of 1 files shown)


deekej:
I have updated to commit to new whitespace/indentation fixes.

sysconfig/network-scripts/network-functions-ipv6
```diff
@@ -1056,10 +1057,28 @@ ipv6_wait_tentative() {
...
         sleep 1
+        countdown=$(($countdown - 1))
     done
-    return 0
+
+    ip_output="$(ip -6 addr show dev ${device} scope global tentative)"
+
+    if [ -n "$ip_output" ]; then
+        net_log $"Some IPv6 address(es) of ${device} remain still in 'tentative' state" info $fn
+        net_log $"Run 'ip -6 addr show dev ${device} scope global tentative' to see more" info $fn
```
lnykryn: There is a return missing in this path.

deekej: I don't get it why it's missing. I thought I had it there... Anyway, I will submit new commit.

lnykryn: Do we really want to fail in that case? Shouldn't we just print warning and go with "success"?

deekej: Thinking about this again after our conversation, do we consider this an error or just a warning?

I'm OK to exit with **success** in case of warning. Otherwise, we should exit with **error**.

lnykryn: I would go with just a warning and return 0.

deekej: OK, so I will update the patch accordingly.

lnykryn:
LGTM

PR #10, opened 2016-10-11
Support boolean values in is_true() & is_false() functions

deekej:
`is_true()` now recognizes value `1` as `true`,
`is_false()` now recognizes value `0` as `false`.

And I replaced the `[yY1]*` and `[nN0]*` occurences with those functions where possible.


deekej:
I have updated the commits to reflect new whitespace/indentation fixes.

lnykryn:
lgtm

PR #11, opened 2016-10-11
[RHEL6] ifup-aliases: do not fail when only IPv6 address is specified

deekej:
Resolves RHBZ [#1340169](https://bugzilla.redhat.com/show_bug.cgi?id=1340169).


lnykryn:
Looks good to me.

PR #16, opened 2016-10-24
Add ability to specify arbitrary route options

jaymzh:
`ifup` can set _some_ route options. Today it's missing several and so, to set
something like `mtu` or `advmss` you have to do hacks like:

```
METRIC='10 mtu 1480'
```

which is sub-optimal. Rather than one-off add support for everything in the man
page, provide a hook for arbitrary extra route options.

### DIFF
--- a/sysconfig/network-scripts/ifup-eth
+++ b/sysconfig/network-scripts/ifup-eth
@@ -284,11 +284,13 @@ else
             # set up default gateway. replace if one already exists
             if [ -n "${GATEWAY}" ] && [ "$(ipcalc --network ${GATEWAY} ${netmask[0]} 2>/dev/null)" = "NETWORK=${NETWORK}" ]; then
                 ip route replace default ${METRIC:+metric $METRIC} \
+                    ${EXTRA_ROUTE_OPTS} \
                     via ${GATEWAY} ${WINDOW:+window $WINDOW} ${SRC} \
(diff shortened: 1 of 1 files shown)


lnykryn:
In other cases we already have MTU variable, maybe do the same thing here so it is not confusing?
Also I would go to just ROUTE_OPTS so it kinda matches with other options.

And we should definitely document those new options somewhere. I guess we should finally rewrite the sysconfig.txt to manpages.

jaymzh:
I didn't want to to be just "MTU" because that would imply *interface* MTU, but this is not, it's *route* MTU. Further we need both MTU and MSS (which is called `adv_mss` on routes), and at that point it seemed silly to be adding two new variables for similar things that will already be confusing with other variables... and then there's the other 30 or so route options that one might need in the future...hence switching to a generic option.

I'm fine with `ROUTE_OPTS` - I went `EXTRA_` just because there's already some route options supported... but I don't mind either way.... is there consensus on that?

I'll add docs to `sysconfig.txt` as well - didn't even know that file existed.

lnykryn:
Hmm you are right in both cases (that is the reson why I wanted to look into it more closely :-) ). Lets merge it as it is now.

lnykryn:
By the way, the sysconfig.txt is extremely outdated in some parts.

PR #19, opened 2016-10-28
netconsole: resolve domain name via 'getent' instead of 'hosts'

deekej:
Adding this fix into RHEL6. For more info, see pull request #18.


deekej:
Commit updated, please see the latest version.

PR #20, opened 2016-10-28
ifup-aliases: don't return with error when arping fails

deekej:
This is a backport of commit 55a50ebc591ebd0f4cfbb8ecc204fa20ee6a7368 to RHEL6, as sugested by Jan Scotka in [RHBZ #1275550](https://bugzilla.redhat.com/show_bug.cgi?id=1275550).


lnykryn:
This will revert part of 37e2084f437172353744b10cad600710137d87c6  and cause regression in https://bugzilla.redhat.com/show_bug.cgi?id=1099026

It looks I have fixed that only on rhel6-branch, I will prepare a pull request for master.

lnykryn:
the code should be close to the one from #26

deekej:
I have updated the commit, to retain the previous changes (to avoid regression).

PR #22, opened 2016-10-30
ifup: do not delegate bridges/bonds/vlans to NetworkManager by default

deekej:
This pull-request solves the issue when the VLANs are managed by NetworkManager,
they can't be brought up. This is a modified patch, originally written by Lubomir Rintel.

Additionally, the Infiniband partitions are ignored as well.

This issue was originally reported in [RHBZ #1284115](https://bugzilla.redhat.com/show_bug.cgi?id=1284115).

### DIFF
--- a/sysconfig/network-scripts/network-functions
+++ b/sysconfig/network-scripts/network-functions
@@ -171,6 +171,19 @@ source_config ()
     else
         ISALIAS=no
     fi
+
+    if ! is_true "$NM_BOND_BRIDGE_VLAN_ENABLED"; then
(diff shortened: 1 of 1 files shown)


lnykryn:
I don't think this is something we want in upstream. NM is now capable of handling those type of devices. Can you check that with Lubomir?

lnykryn:
I don't think we need this in upstream. Current NM is capable of handling VLANs, bonds,....

deekej:
OK, so according to Lubomir, this should remaina a downstream patch, if needed.

PR #23, opened 2016-10-30
ifup: do not delegate bridges/bonds/vlans to NetworkManager by default

deekej:
Adding this fix into RHEL6. For more info, see pull-request #22.


deekej:
So, do you suggest to rework this patch, or should we close this as WONTFIX?

lnykryn:
According to NM maintainer we should leave the code as is.

PR #24, opened 2016-10-30
ifup-aliases: send gratuitous ARPs when adding addresses

deekej:
Excerpt from [RHBZ #1320366](https://bugzilla.redhat.com/show_bug.cgi?id=1320366):

_`ifup-aliases` does not send gratuitous ARP updates when adding addresses.
When moving one or more aliases from one server to another, those IPs may not be
reachable for some time (depending on the ARP timeout on the upstream router)._

### DIFF
--- a/sysconfig/network-scripts/ifup-aliases
+++ b/sysconfig/network-scripts/ifup-aliases
@@ -277,6 +277,12 @@ function new_interface ()
         /sbin/ip addr add ${IPADDR}/${PREFIX} brd ${BROADCAST} \
             dev ${parent_device} label ${DEVICE}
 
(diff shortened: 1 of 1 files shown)


deekej:
The patch itself looks good to me, but I'm not sure about the 2 seconds `sleep` used there. IMHO, the switches/routers nowadays are quite fast, even for bigger enterprises. I think 1 second should be sufficent, and we could consider sleeping for even less time. What do you think, guys?

lnykryn:
The code is the same in ifup-eth. And since it runs in background, I don't think it is a problem.

lnykryn:
LGTM

PR #25, opened 2016-10-31
ifup-aliases: don't return with error when arping fails

deekej:
Backport of commit 55a50ebc591ebd0f4cfbb8ecc2, to resolve [RHBZ #1333544](https://bugzilla.redhat.com/show_bug.cgi?id=1333544).


deekej:
OK, I missed this because I have done this yesterday, but this is duplicate of #20.

PR #27, opened 2016-10-31
network: don't put down network when /usr is remote-fs

lnykryn:
Nowadays /proc is a essential part of system and if it is on network,
we need to unmount it in shutdown dracut.

Also parsing /etc/mtab does not work anymore.

Fixes: rhbz#1369790


lnykryn:
Yep that was a typo, I have corrected the commit message. And yes if the /usr is on separate partion it would it mounted and unmounted in dracut.

PR #33, opened 2016-11-09
init/kexec-disable.conf: release kexec memory properly

deekej:
Previously, the kexec memory was freed everytime. After patch from [RHBZ #734987](https://bugzilla.redhat.com/show_bug.cgi?id=734987),
it does not free the kexec memory at all.

This patch fixes this regression (resolves [RHBZ #1378198](https://bugzilla.redhat.com/show_bug.cgi?id=1378198)).


deekej:
The commit has been fixed.

PR #35, opened 2016-11-10
network-functions: do not send hostname via dhclient everytime

deekej:
The commit cd475bf315b9c08e9f1732c0b9de95729025d1c5 introduced a regression where dhclient is now sending the current hostname (the `-H` option) everytime when `ifup-eth` is used.

This can cause a problem in case the hostname is not specified in `/etc/sysconfig/network` or `/etc/sysconfig/network-scripts/ifcfg-eth*`, where the dhclient would eventually send `localhost` as current hostname, which is obviously wrong.

This PR tries to fix this issue by checking if the hostname is set or not.

Originally reported in [RHBZ #1350602](https://bugzilla.redhat.com/show_bug.cgi?id=1350602).

I'm not sure how `dhclient` behaves in case no hostname is specified, so the review of this commit is especially needed.

### DIFF
--- a/sysconfig/network-scripts/ifup-eth
+++ b/sysconfig/network-scripts/ifup-eth
@@ -185,7 +185,13 @@ if [ -n "${DYNCONFIG}" ] && [ -x /sbin/dhclient ]; then
     fi;
     generate_config_file_name
     generate_lease_file_name
-    DHCLIENTARGS="${DHCLIENTARGS} -H ${DHCP_HOSTNAME:-${HOSTNAME%%.*}} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
+
+    if need_hostname; then
+        DHCLIENTARGS="${DHCLIENTARGS} ${DHCP_HOSTNAME:+-H $DHCP_HOSTNAME} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
+    else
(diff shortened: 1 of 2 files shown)


sysconfig/network-scripts/ifup-eth
```diff
@@ -326,7 +332,14 @@ if is_true "${DHCPV6C}" && [ -x /sbin/dhclient ]; then
     generate_lease_file_name 6
     echo
     echo -n $"Determining IPv6 information for ${DEVICE}..."
-    if /sbin/dhclient -6 -1 ${DHCPV6C_OPTIONS} ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient6-${DEVICE}.pid -H ${DHCP_HOSTNAME:-${HOSTNAME%%.*}} ${DEVICE} ; then
+
+    if need_hostname; then
+        DHCLIENTARGS="-6 -1 ${DHCPV6C_OPTIONS} ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient6-${DEVICE}.pid ${DEVICE}"
```
lnykryn: DHCLIENTARGS="-6 -1 ${DHCPV6C_OPTIONS} ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient6-${DEVICE}.pid ${DEVICE}" ->
DHCLIENTARGS="-6 -1 ${DHCPV6C_OPTIONS} ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient6-${DEVICE}.pid ${DHCP_HOSTNAME:+-H $DHCP_HOSTNAME} ${DEVICE}"

deekej:
Clearing the wrong labels and milestones. I will create a separate PR for RHEL6 backport.

PR #39, opened 2016-11-10
sysconfig/network: NO_DHCP_HOSTNAME option introduced

deekej:
Previously, some administrators were unable to force initscripts to *NOT* obtain hostname from DHCP, even though they were using static configuration of network. Righ now, setting  `NO_DHCP_HOSTNAME` to `yes`, `true` or `1` in `/etc/sysconfig/network` will allow them to do so.

### DIFF
--- a/sysconfig/network-scripts/ifup-eth
+++ b/sysconfig/network-scripts/ifup-eth
@@ -186,10 +186,12 @@ if [ -n "${DYNCONFIG}" ] && [ -x /sbin/dhclient ]; then
     generate_config_file_name
     generate_lease_file_name
 
-    if need_hostname; then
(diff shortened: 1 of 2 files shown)


deekej:
NOTE: This needs to be added to RHEL6 as well, after this has been merged.

deekej:
Actually, the change will need to be done in `sysconfig/network-scripts/ifup-eth` file (commit e20a9baff26986d2bbb8201615dd39c9748b7141), because there is a 3rd use case that needs to be handled there. I guess I will prepare it as part of this PR.

deekej:
I have updated the commit. It introduces a new function `need_hostname_check()`, but I'm not sure about the name. If you can think of something better, please let me know.

lnykryn:
Instead of need_hostname_check I would use something like "is_hostname_set"

deekej:
Yeah, you're right actually, I forgot about that line... o.O I was thinking about `is_hostname_set` before, but this would also require some change in the logic.

deekej:
Commit updated.

PR #47, opened 2016-11-21
Allow someone to keep accept_ra on while using static addresses.

deekej:
```
It's reasonable (and common) to assign static addresses while still
wanting to get your default gw from your router. The assumption that not
wanting SLAAC means not wanting RA is flawed, but I don't want to break
backwards compatability, so I'm adding an additional option.
```

This is backport of this option in RHEL6. Related to [RHBZ #1086388](https://bugzilla.redhat.com/show_bug.cgi?id=1086388).


deekej:
This can't be backported - it would mean changing default behaviour for our customers - as already discussed with @lnykryn.

PR #57, opened 2017-01-31
use dbus instead of nmcli to make ifup/ifdown faster

phoracek:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1408219

Kudos Beniamino Galvani

### DIFF
--- a/sysconfig/network-scripts/network-functions
+++ b/sysconfig/network-scripts/network-functions
@@ -172,12 +172,20 @@ source_config ()
         ISALIAS=no
     fi
     if is_nm_running && [ "$REALDEVICE" != "lo" ] ; then
-        nmcli con load "/etc/sysconfig/network-scripts/$CONFIG"
+        nm_con_load "$CONFIG"
         if ! is_false $NM_CONTROLLED; then
             UUID=$(get_uuid_by_config $CONFIG)
             [ -n "$UUID" ] && _use_nm=true
         fi
     fi
+}
+
+nm_con_load () {
+    dbus-send --system --print-reply \
+        --dest=org.freedesktop.NetworkManager \
+        /org/freedesktop/NetworkManager/Settings \
+        org.freedesktop.NetworkManager.Settings.LoadConnections \
+        array:string:"/etc/sysconfig/network-scripts/${1}" >/dev/null 2>&1
 }
 
 ethtool_set()
--- a/sysconfig/network-scripts/network-functions
+++ b/sysconfig/network-scripts/network-functions
@@ -277,7 +277,11 @@ do_netreport ()
 
 is_nm_running ()
 {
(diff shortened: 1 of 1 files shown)


sysconfig/network-scripts/network-functions
```diff
@@ -172,14 +172,22 @@ source_config ()
         ISALIAS=no
     fi
     if is_nm_running && [ "$REALDEVICE" != "lo" ] ; then
-        nmcli con load "/etc/sysconfig/network-scripts/$CONFIG"
+        nm_con_load $CONFIG
```
thom311: I would quote $CONFIG (although there are other places where quotation is missing)

sysconfig/network-scripts/network-functions
```diff
@@ -172,14 +172,22 @@ source_config ()
...
         fi
     fi
 }
 
+nm_con_load () {
+    dbus-send --system --print-reply \
+        --dest=org.freedesktop.NetworkManager \
+        /org/freedesktop/NetworkManager/Settings \
+        org.freedesktop.NetworkManager.Settings.LoadConnections \
+        array:string:"/etc/sysconfig/network-scripts/$1"
```
thom311: I think this should redirect the output to /dev/null to avoid spamming stdout/stderr

dankenigsberg:
@deekej , can you elaborate why this does not fit to master?

(personally I can live with this in el7 only, but the slowliness exists in Fedora as well, afaik)

phoracek:
@thom311 thanks for your comments, it should be fixed now.

sysconfig/network-scripts/network-functions
```diff
@@ -172,14 +172,22 @@ source_config ()
         ISALIAS=no
     fi
     if is_nm_running && [ "$REALDEVICE" != "lo" ] ; then
-        nmcli con load "/etc/sysconfig/network-scripts/$CONFIG"
+        nm_con_load ${CONFIG}
```
thom311: This is the same as plain `$CONFIG`. I meant adding quoting like `"$CONFIG"`. Anyway, as you prefer.

deekej:
Sorry guys for the delay, I got caught in some other urgent matters... :-/

Anyway, because we already have another call to D-Bus in initscripts, the SELinux policy should be okay with that, AFAIK. And if it will also mean a little bit faster start of network for Fedora (in some cases), I don't have anything more against it.

I realize it might not be the cleanest solution, but in the spirit of Zen of Python `practicality beats purity` I will accept these changes. :) (I just want to look at the code again after latest commits. Hopefully, this will be in Fedora at the end of this week.)

phoracek:
Thanks @thom311, fixed. @deekej I changed `nmcli_con_load` call a bit just now, hope it's ok, thank you.

deekej:
Thank you for your help, Lukas.

PR #63, opened 2017-03-01
po: download latest translations from zanata

lnykryn:


piotrdrag:
FWIW, it looks correct to me.

lnykryn:
Just a small note. The huge amnout of removed lines is caused by the fact, that in the past initscripts package was shipping translation for every initscript on the system (bind, apache, ...). Now everything is migrated to systemd unit files.

PR #64, opened 2017-03-05
ifup can create veth devices

lvgilmore:
minor change to enable veth ifcfg files.
been using it in my environment for some time and found it helpful, maybe some else will too.

first pr to public repo, so if i''ve skipped a required procedure please let me know.

### DIFF
--- a/sysconfig/network-scripts/ifdown-eth
+++ b/sysconfig/network-scripts/ifdown-eth
@@ -178,4 +178,10 @@ if [ -n "$VLAN" ]; then
     fi
 fi
 
+if [ "${TYPE}" = "Veth" ]; then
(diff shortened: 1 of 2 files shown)


lvgilmore:
Thanks for the input!
I'm wondering what is the desired effect of ifdown on veth device. Is it possible that the user may want to set one peer of the pair down while keeping the other up?
Right now I've edited ifdown to delete the pair only if both peers are down

jamacku (commented):
LGTM

jamacku:
> Right now I've edited ifdown to delete the pair only if both peers are down

Thanks for your contribution, I think this is the correct approach.

PR #67, opened 2017-03-21
[RHEL-7] 9.70 Sync

deekej:
We have decided to synchronize the network scripts (and other) functionality with Fedora initscripts without removing the necessary parts from RHEL-7 (for backward compatibility).

Basically, I have copied the files from 9.70 into rhel7-branch, and created separate commits for it. The proper review will be needed.

NOTE: This is just a PR to bring RHEL-7 in sync with upstream as much as we can. There are still some other issues that needs fixing both in upstream & RHEL-7.


deekej:
I have added @jaymzh to this in case he would like to help us. He already has experience with backporting the upstream initscripts into RHEL-7, and I could have overlook something.

rc.d/init.d/network
```diff
@@ -2,7 +2,7 @@
 #
 # network       Bring up/down networking
 #
-# chkconfig: 2345 10 90
+# chkconfig: - 10 90
```
lnykryn: Change in this line will cause that on new installed systems network initscript will be disabled by default. We definitely don't want that!

jaymzh:
👍  sorry didn't get around to this one yesterday, but it looks good.

deekej:
No problem, Phil! :) But thanks for checking it! ;)

deekej:
Accidentally I have returned the regression of dhclient arguments being passed as one parameter. #81 solves the problem.

PR #89, opened 2017-03-29
Optionally do not send hostname with dhclient -H

robbmanes:
When using DHCP and invoking dhclient, ifup-eth always passes the -H
flag which provides a hostname to the DHCP server.  On systems with
multiple interfaces it is sometimes required for only one interface to
send a hostname.

This patch adds the DHCP_SEND_HOSTNAME and the DHCP6_SEND_HOSTNAME
options which, when set to "no" or "off", remove the -H from the
dhclient command.  These are added to the user's ifcfg-$INTERFACE file.

### DIFF
--- a/sysconfig/network-scripts/ifup-eth
+++ b/sysconfig/network-scripts/ifup-eth
@@ -186,12 +186,17 @@ if [ -n "${DYNCONFIG}" ] && [ -x /sbin/dhclient ]; then
     generate_config_file_name
     generate_lease_file_name
 
-    if is_hostname_set; then
-        # We already have the hostname ->> send it to DHCP:
-        DHCLIENTARGS="${DHCLIENTARGS} -H ${DHCP_HOSTNAME:-${HOSTNAME%%.*}} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
+    # Check to see if we should pass -H to dhclient
+    if [ "${DHCP_SEND_HOSTNAME}" = "no" -o "${DHCP_SEND_HOSTNAME}" = "off" ]; then
+        DHCLIENTARGS="${DHCLIENTARGS} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
     else
-        # We need to acquire the hostname:
-        DHCLIENTARGS="${DHCLIENTARGS} ${DHCP_HOSTNAME:+-H $DHCP_HOSTNAME} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
+        if is_hostname_set; then
+            # We already have the hostname ->> send it to DHCP:
+            DHCLIENTARGS="${DHCLIENTARGS} -H ${DHCP_HOSTNAME:-${HOSTNAME%%.*}} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
+        else
+            # We need to acquire the hostname:
+            DHCLIENTARGS="${DHCLIENTARGS} ${DHCP_HOSTNAME:+-H $DHCP_HOSTNAME} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
+        fi
     fi
 
     echo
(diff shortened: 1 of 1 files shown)


robbmanes:
By default we pass '-H' to dhclient to send hostnames, but not all DHCP servers require this.  In some cases, this is not desired.

Without the DHCP_SEND_HOSTNAME flag (or if it is set to anything else):

# cat ifcfg-ens3 
HWADDR=52:54:00:4F:A6:78
TYPE=Ethernet
BOOTPROTO=dhcp
ONBOOT=yes
AUTOCONNECT_PRIORITY=-999
NM_CONTROLLED=no

# ps aux | grep dhclient
root      4201  0.0  0.6  87132 12320 ?        Ss   15:54   0:00 /sbin/dhclient -H fedora25-server -1 -q -lf /var/lib/dhclient/dhclient-fcb6d9a1-5256-3dcb-9243-9cda7cf0a953-ens3.lease -pf /var/run/dhclient-ens3.pid ens3

And with the flag set to "no":

# cat ifcfg-ens3 
HWADDR=52:54:00:4F:A6:78
TYPE=Ethernet
BOOTPROTO=dhcp
ONBOOT=yes
AUTOCONNECT_PRIORITY=-999
DHCP_SEND_HOSTNAME=no

# ps aux | grep dhclient
root      4620  0.0  0.6  87132 12316 ?        Ss   15:58   0:00 /sbin/dhclient -1 -q -lf /var/lib/dhclient/dhclient-fcb6d9a1-5256-3dcb-9243-9cda7cf0a953-ens3.lease -pf /var/run/dhclient-ens3.pid ens3

In addition, we add the DHCP6_SEND_HOSTNAME flag for IPv6.

Let me know if there needs to be any corrections or addendum.  Thanks.

deekej:
Hello, sorry for the late reply, I was engaged in my other responsibilities... Anyway, lets get to it:

So, looking at your description and commit, I get the feeling that you want to obtain the hostame for your machine from DHCP. (In other words, specifying -H to dhclient tells it to send the specified hostname to DHCP server, and not specifying it at all will use the received hostname from DHCP.)

However, we already have this functionality in initscripts already (see https://github.com/fedora-sysv/initscripts/blob/master/sysconfig/network-scripts/ifup-eth#L193):

deekej:
Hello, sorry for the late reply, I was engaged in my other responsibilities... Anyway, lets get to it:

So, looking at your description and commit, I get the feeling that you want to obtain the hostame for your machine from DHCP. (In other words, specifying -H to dhclient tells it to send the specified hostname to DHCP server, and not specifying it at all will use the received hostname from DHCP.)

However, we already have this functionality in initscripts already (see [here](https://github.com/fedora-sysv/initscripts/blob/master/sysconfig/network-scripts/ifup-eth#L189)):

```bash
if is_hostname_set; then
    # We already have the hostname ->> send it to DHCP:
    DHCLIENTARGS="${DHCLIENTARGS} -H ${DHCP_HOSTNAME:-${HOSTNAME%%.*}} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
else
    # We need to acquire the hostname:
    DHCLIENTARGS="${DHCLIENTARGS} ${DHCP_HOSTNAME:+-H $DHCP_HOSTNAME} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
fi
```

In case we already have hostname set (and it's something valid - not `(none)`, `localhost` or `localhost.localdomain`), the we send that hostname to DHCP. It is not expected that you would like to set machine's hostname to have it rewritten by DHCP. [This is the if statement.]

In case the hostname is not set (or it is `(none)`, `localhost` or `localhost.localdomain`), then it is expected you want to obtain your hostname from DHCP. If the `DHCP_HOSTNAME` is not set, then the `-H` option will never be used for dhclient.

In other words - if you don't want to use`-H` with dhclient in `ifup-eth` script, then you shouldn't have set the `DHCP_HOSTNAME` configuration option. IMHO, this is logical. It does not make much sense to ask network scripts to set the hostname for you with `DHCP_HOSTNAME` option and ask to not use `-H` option (i.e. to not set the hostname) at the same time. Either you can do option A, or you can do option B - but you can't both at the same time... ;)

Or maybe I'm not understanding your use-case properly. For now, I'm closing this pull-request, but feel free to reopen it if you have a valid use-case for this. :)

deekej:
Hello @robbmanes,

because I was doing some major rework of a way how initscripts currently handle the DHCP_HOSTNAME option, I was able to incorporate your suggestion for DHCP_SEND_HOSTNAME into it as well. Please, sell pull-request #92 for more information. I hope this will help your needs. ;)

Best regards,

Dee'Kej

robbmanes:
> Please, sell pull-request #92 for more information. I hope this will help your needs. ;)

Sorry about the delay - I think it does.  Thank you very much!

PR #92, opened 2017-04-20
ifup-eth: introduce DHCP_FQDN option

deekej:
The `DHCP_FQDN` is similar to `DHCP_HOSTNAME` option, but it expects the fully qualified domain name to be used there.

In case both `DHCP_HOSTNAME` and `DHCP_FQDN` are set, then the `DHCP_FQDN` will used (*and warning will be produced*). This is the same behaviour as *NetworkManager* does it.

### DIFF
--- a/sysconfig/network-scripts/ifup-eth
+++ b/sysconfig/network-scripts/ifup-eth
@@ -186,13 +186,9 @@ if [ -n "${DYNCONFIG}" ] && [ -x /sbin/dhclient ]; then
     generate_config_file_name
     generate_lease_file_name
 
-    if is_hostname_set; then
-        # We already have the hostname ->> send it to DHCP:
-        DHCLIENTARGS="${DHCLIENTARGS} -H ${DHCP_HOSTNAME:-${HOSTNAME%%.*}} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
-    else
-        # We need to acquire the hostname:
-        DHCLIENTARGS="${DHCLIENTARGS} ${DHCP_HOSTNAME:+-H $DHCP_HOSTNAME} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
-    fi
+    # Initialize the dhclient args and obtain the hostname options if needed:
+    DHCLIENTARGS="${DHCLIENTARGS} ${ONESHOT} -q ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient-${DEVICE}.pid"
+    set_hostname_options DHCLIENTARGS
 
     echo
     echo -n $"Determining IP information for ${DEVICE}..."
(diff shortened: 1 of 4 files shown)


sysconfig/network-scripts/ifup-eth
```diff
@@ -335,12 +340,17 @@ if is_true "${DHCPV6C}" && [ -x /sbin/dhclient ]; then
...
 
+    # Get the correct option to be sent to dhclient:
+    HOSTNAME_PARAM=''
+    HOSTNAME_VALUE=''
+    set_dhcp_hostname_options HOSTNAME_PARAM HOSTNAME_VALUE
+
     if is_hostname_set; then
         # We already have the hostname ->> send it to DHCP:
-        DHCLIENTARGS="-6 -1 ${DHCPV6C_OPTIONS} ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient6-${DEVICE}.pid -H ${DHCP_HOSTNAME:-${HOSTNAME%%.*}} ${DEVICE}"
+        DHCLIENTARGS="-6 -1 ${DHCPV6C_OPTIONS} ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient6-${DEVICE}.pid ${HOSTNAME_PARAM} ${HOSTNAME_VALUE:-${HOSTNAME%%.*}} ${DEVICE}"
```
lnykryn: What if neither DHCP_HOSTNAME nor DHCP_FQDN is set? Then ${HOSTNAME_PARAM} will be empty, but ${HOSTNAME_VALUE:-${HOSTNAME%%.*}} be expanded as HOSTNAME

sysconfig/network-scripts/network-functions
```diff
@@ -365,6 +366,26 @@ is_available_wait ()
...
+{
+    # DHCP_FQDN takes precedence before DHCP_HOSTNAME:
+    if [ -n "${DHCP_FQDN}" ]; then
+        eval "$1='-F'"
+        eval "$2=${DHCP_FQDN}"
+    else
+        eval "$1='-H'"
+        eval "$2=${DHCP_HOSTNAME}"
+    fi
+
```
lnykryn: Is there a specific reason why we have to use to separate variables and eval? Also in the old code we are using HOSTNAME if dhcp hostname is not used. I think we should move that to this function as well. That would also mean that we could remove those two  "if is_hostname_set; then"
can't we just do something like this:

set_dhcp_hostname_options
DHCLIENTARGS="-6 -1 ${DHCPV6C_OPTIONS} ${DHCLIENTCONF} -lf ${LEASEFILE} -pf /var/run/dhclient6-${DEVICE}.pid ${DHCP_HOSTNAME} ${DEVICE}"


set_dhcp_hostname_options () {
    DHCP_HOSTNAME_OPTIONS=""
    if [ -n "${DHCP_FQDN}" ]; then
        DHCP_HOSTNAME_OPTIONS="-F ${DHCP_FQDN}"
    else if [ -n "${DHCP_HOSTNAME}" ]; then 
        DHCP_HOSTNAME_OPTIONS="-H ${DHCP_HOSTNAME}"
    else  if is_hostname_set; then
        DHCP_HOSTNAME_OPTIONS="-H ${HOSTNAME%%.*}"
    fi
........

deekej:
I have done complete rework of the DHCP_FQDN option. Please see the commit message and comments inside the commit for more info. Thank you!

deekej:
Self-note: Unfortunately, `dhclient-script` stopped sourcing the `sysconfig/network-scripts/network-functions` in the Fedora 25, which might lead to regressions. Maybe create a new issue to unite the behaviour again?

deekej:
> Besides my one nitpick it looks fine.
Yeah, you were right. I have removed that redundant part.

PR #94, opened 2017-04-25
init.d/network: do not bring network down when /usr is on iSCSI

deekej:
The current solution is not sufficient. The system would still hang when `/usr` is located on iSCSI devide.

This commit brings back the previous solution, but instead of looking into `/etc/mtab` it parses the `/etc/fstab` file.

This was the idea of @msekletar.

### DIFF
--- a/rc.d/init.d/network
+++ b/rc.d/init.d/network
@@ -84,16 +84,22 @@ start)
 
     # bring up all other interfaces configured to come up at boot time
     for i in $interfaces; do
-        unset DEVICE TYPE SLAVE NM_CONTROLLED
+        unset DEVICE TYPE SLAVE NM_CONTROLLED NETBOOT
         eval $(LANG=C grep -F "DEVICE=" ifcfg-$i)
         eval $(LANG=C grep -F "TYPE=" ifcfg-$i)
         eval $(LANG=C grep -F "SLAVE=" ifcfg-$i)
         eval $(LANG=C grep -F "NM_CONTROLLED=" ifcfg-$i)
+        eval $(LANG=C grep -F "NETBOOT=" ifcfg-$i)
 
         if [ -z "$DEVICE" ] ; then DEVICE="$i"; fi
 
         if [ "$SLAVE" = "yes" ] && ( ! is_nm_running || is_false $NM_CONTROLLED ) ; then
             continue
+        fi
+
+        # NOTE: Initialization of NETBOOT-enabled interface must be done manually with ifup script.
+        if is_true "${NETBOOT}"; then
+          continue
         fi
 
         if [ "${DEVICE##cipcb}" != "$DEVICE" ] ; then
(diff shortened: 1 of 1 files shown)


msekletar:
> The current solution is not sufficient. The system would still hang when /usr is located on iSCSI devide.

Do we have a clear understanding why current solution is not sufficient when ```/usr``` is located on iSCSI disk? In such case mount should be marked with ```_netdev``` option in ```/etc/fstab``` and corresponding systemd mount unit *should* have a dependency on ```remote-fs.target```.

> This commit brings back the previous solution, but instead of looking into /etc/mtab it parses the /etc/fstab file.

This is not sufficient either, there are other *network* filesystems except NFS. 

> This was the idea of @msekletar.

I don't recall where exactly I proposed this terrible awk expressions, but I'd like to make clear that they are very bad idea (and were back then as well, but I was being silly I guess).

deekej:
> I don't recall where exactly I proposed these terrible awk expressions, but I'd like to make clear that they are very bad idea (and were back then as well, but I was being silly I guess).
I have derived this commit from your [BZ #1369790 comment 2 ](https://bugzilla.redhat.com/show_bug.cgi?id=1369790#c2).

> Do we have a clear understanding why current solution is not sufficient when /usr is located on iSCSI disk? In such case mount should be marked with _netdev option in /etc/fstab and corresponding systemd mount unit should have a dependency on remote-fs.target.

So, the current solution might work in Fedora (I haven't tested it there), but it does not work in RHEL-7.3. (Probably because of older systemd?) If I run `systemctl show --property=RequiredBy -- usr.mount` I will get `initrd-root-fs.target` and NOT the `remote-fs.target`.

```bash
[root@rhevh-x3650-01-g8 ~]# systemctl show --property=RequiredBy -- usr.mount
RequiredBy=initrd-root-fs.target
```
So without the test for `usropts` the system will still hang during the shutdown, because the network was brought down too soon.

> This is not sufficient either, there are other network filesystems except NFS.
I see. Do you have any idea how to make this better?

msekletar:
I am starting to think that this doesn't need any fix in initscripts after all, but in systemd. Retaining dependency on ```initrd-root-fs.target``` is not a problem, but the fact that filesystem marked with ```_netdev``` option doesn't depend on ```remote-fs.target``` is.

```
[root@rhevh-x3650-01-g8 ~]# systemctl show usr.mount | grep remote-fs
[root@rhevh-x3650-01-g8 ~]# systemctl show usr.mount | grep local-fs
Before=initrd-root-fs.target local-fs.target
```

deekej:
> I am starting to think that this doesn't need any fix in initscripts after all, but in systemd.

OK - should I open a new RHBZ for it?

Anyway, we had a discussion with @lnykryn and @pholica about this, and we have come up with an enhancement for `/etc/init.d/network` - it shouldn't do anything with network interfaces (config files) containing `NETBOOT=yes`. Dracut puts `NETBOOT=yes` into ifcfg files, but it has not been used so far. We can use this as indication when we should not automatically bring up/down these interfaces.

IMHO, this should work correctly even if you fix the systemd issue.

deekej:
I have replaced the commit with different changes, based on our previous discussion.

deekej:
So, after further discussion of *systemd* and *dracut* developers, it was decided to keep this solution, and the fix will be done in *dracut*, which should make [this line](https://github.com/fedora-sysv/initscripts/blob/master/rc.d/init.d/network#L165) work for `/usr` as well.

More info in [RHBZ #1446171](https://bugzilla.redhat.com/show_bug.cgi?id=1446171).

PR #102, opened 2017-05-25
Replace usleep(1) calls with sleep(1) calls

deekej:
Also, add the deprecation warning into manual page of `usleep(1)`.

### DIFF
--- a/sysconfig/network-scripts/network-functions
+++ b/sysconfig/network-scripts/network-functions
@@ -196,7 +196,11 @@ ethtool_set()
 {
     oldifs=$IFS;
     IFS=';';
-    [ -n "${ETHTOOL_DELAY}" ] && /bin/usleep ${ETHTOOL_DELAY}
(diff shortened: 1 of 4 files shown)


sysconfig/network-scripts/network-functions
```diff
@@ -196,7 +196,7 @@ ethtool_set()
 {
     oldifs=$IFS;
     IFS=';';
-    [ -n "${ETHTOOL_DELAY}" ] && /bin/usleep ${ETHTOOL_DELAY}
+    [ -n "${ETHTOOL_DELAY}" ] && /usr/bin/sleep ${ETHTOOL_DELAY}
```
lnykryn: Hmm this does not look equivalent. Old code takes the setting in usec and the new one in sec

deekej: Yes, you're correct. I will fix this.

PR #111, opened 2017-07-21
Drop sysctl.d/00-system.conf

keszybz:
This file now only contains lines to disable netfilter on bridges. In kernel 3.18 this filtering functionality was made non-default by moving it to br_netfilter [BZ #512206](https://bugzilla.redhat.com/show_bug.cgi?id=512206). Anybody who actually wants to use `br_netfilter` has to load it explicitly anyway, so disabling it through sysctl isn't necessary anymore.

The reason for removal is that by default (i.e. when `br_netfilter` is not loaded), we get a warning on every boot (*which is confusing for users and inelegant*):
``` log
systemd-sysctl[210]: Couldn't write '0' to 'net/bridge/bridge-nf-call-ip6tables', ignoring: No such file or directory
systemd-sysctl[210]: Couldn't write '0' to 'net/bridge/bridge-nf-call-iptables', ignoring: No such file or directory
systemd-sysctl[210]: Couldn't write '0' to 'net/bridge/bridge-nf-call-arptables', ignoring: No such file or directory
```
The downside of removing this file is for people who load `br_netfilter` for some reason *and* do no want to use it, will have to take an additional step now (either restore the sysctl settings or remove `br_netfilter` from `/etc/modules-load.d` or wherever). *I expect the number of people affected to be very small.*

(Note that the file was overwritten on sparc and s390, so those architectures see no change.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,13 +26,10 @@ install:
 	install -m644  debug.csh debug.sh $(ROOT)/etc/profile.d
 	install -m755  sys-unconfig $(ROOT)/usr/sbin
 	install -m644  service.8 sys-unconfig.8 $(ROOT)$(mandir)/man8
-	mkdir -p -m 755 $(ROOT)/usr/lib/sysctl.d
-	mkdir -p -m 755 $(ROOT)/etc/sysctl.d
-	install -m644 sysctl.conf $(ROOT)/usr/lib/sysctl.d/00-system.conf
 	if uname -m | grep -q sparc ; then \
-	  install -m644 sysctl.conf.sparc $(ROOT)/usr/lib/sysctl.d/00-system.conf ; fi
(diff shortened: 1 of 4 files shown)


keszybz:
systemd already has a bunch of sysctl files, so there's no need to keep 00-system.conf around. In particular, initscripts are optional (or at least should be), so it's not appropriate to keep *any* sysctl files in here that have a general purpose.

I didn't touch `sysctl.conf.{s390,sparc}` because I know very little about those architectures. But the same principle applies: most likely those settings should be move to systemd or dropped.

lnykryn:
I would love to drop all sysctl configuration, but I am but worried that our rhel users will see it as a regression in next major version of rhel.

@deekej I am fine with merging this, but please make sure that we will release a knowledge base article about this, when time is right.

keszybz:
Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=1264588.

PR #112, opened 2017-07-21
[Bridge] Wait for STP to complete setup

deekej:
This pull-request contains 2 commits. First introduces an auxiliary function for converting values of time units smaller than 1 second into seconds. The second fixes an issue when bridge can't obtain IP from DHCP right after system start, because `ifup-eth` does not wait for STP to complete its starup ([RHBZ #1380496](https://bugzilla.redhat.com/show_bug.cgi?id=1380496)).

The second commit is based on patch from Richard Alloway (@N3WWN).

### DIFF
--- a/sysconfig/network-scripts/ifup-eth
+++ b/sysconfig/network-scripts/ifup-eth
@@ -51,13 +51,16 @@ if [ "${TYPE}" = "Bridge" ]; then
         net_log $"Bridge support not available: brctl not found"
         exit 1
     fi
+
     if [ ! -d /sys/class/net/${DEVICE}/bridge ]; then
         /usr/sbin/brctl addbr -- ${DEVICE} || exit 1
     fi
+
     [ -n "${DELAY}" ] && /usr/sbin/brctl setfd -- ${DEVICE} ${DELAY}
     [ -n "${STP}" ] && /usr/sbin/brctl stp -- ${DEVICE} ${STP}
     [ -n "${PRIO}" ] && /usr/sbin/brctl setbridgeprio ${DEVICE} ${PRIO}
     [ -n "${AGEING}" ] && /usr/sbin/brctl setageing ${DEVICE} ${AGEING}
+
     # add the bits to setup driver parameters here
     for arg in $BRIDGING_OPTS ; do
         key=${arg%%=*};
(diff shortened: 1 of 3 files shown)


sysconfig/network-scripts/ifup-eth
```diff
@@ -54,7 +54,17 @@ if [ "${TYPE}" = "Bridge" ]; then
...
     fi
-    [ -n "${DELAY}" ] && /usr/sbin/brctl setfd -- ${DEVICE} ${DELAY}
+
+    # Set the DELAY as forward_delay to bridge, if set. Otherwise, set the DELAY to forward_delay value,
+    # so the STP can correctly complete the startup first before trying to obtain IP address from DHCP.
+    if [[ -n "${DELAY}" ]]; then
+      /usr/sbin/brctl setfd -- ${DEVICE} ${DELAY}
+    else
+      forward_delay="$(cat /sys/devices/virtual/net/${DEVICE}/bridge/forward_delay)"
+      DELAY="$(convert2sec ${forward_delay} centi)"
```
lnykryn: Huh? That time is rellay in centiseconds?

deekej: Yes, it is. I was surprised myself as well when I saw the original patch from @N3WWN for the first time.

In case you have some virtual bridge device on your machine, you can check this yourself by following these steps:
1) Display the internal kernel value for `forward_delay` for given bridge device:
```
~ $ cat /sys/devices/virtual/net/virbr0/bridge/forward_delay
200
```
2) Now, use `brctl` to display information about `virbr0`:
```
~ $ brctl showstp virbr0 
virbr0
 bridge id		8000.525400bad86e
 designated root	8000.525400bad86e
 root port		   0			path cost		   0
 max age		  20.00			bridge max age		  20.00
 hello time		   2.00			bridge hello time	   2.00
 forward delay		   2.00			bridge forward delay	   2.00
 ageing time		 300.00
 hello timer		   1.28			tcn timer		   0.00
 topology change timer	   0.00			gc timer		   0.00
 flags			

(... rest of output ommited)	
```
Here you can see the `forward delay` and `bridge forward delay` are both `2.00`. According to man page for `brctl (8)`:
```man
brctl setfd <bridge> <time> sets the bridge's 'bridge forward delay' to <time> seconds.
```
(You can try to change the bridge forward delay value to see that e.g. 5 seconds will be displayed as 5.00 via 'brctl showstp`.)

Therefore, if I get back to the example above... `2.00 seconds` equal value `200` (as represented in kernel). Thus:
`2.00 sec == 200 centisec`

I hope my logical deduction makes sense... :D

deekej:
> Hmm, if this is related just to the linkdelay shouldn't we move this piece of code few lines below to the
"# set LINKDELAY.... if ..." part?

IMHO, this is just a cosmetic issue. As you say, it makes more sense to have it there (in # set LINKDELAY section), and I was thinking of that initially. But that section felt already complicated enough before, so I didn't want to add more "obfuscation" on top it, and possibly thus creating new regression.

If you insist, I can move it to that part. Personally I find this more clean. :)

deekej:
I have changed the commit based on our discussion. It has been moved to section where we set the `LINKDELAY`, and the `DELAY ` variable is not being changed anymore, instead it uses temporary variable.

I have also added a new commit for `is_true()` and `is_false()` to support `ON` & `OFF` values.

PR #127, opened 2017-09-19
Remove symlinks for fedora-* services

deekej:
These symlinks are no longer needed, because services enabled by default are now managed by `fedora-release` package.

Before merging this PR/releasing new version of initscripts, we should wait for necessary changes to be made in `fedora-release`.

### DIFF
--- a/initscripts.spec
+++ b/initscripts.spec
@@ -29,6 +29,7 @@ Conflicts: ppp < 2.4.6-4
 Requires(pre): /usr/sbin/groupadd
 Requires(post): /sbin/chkconfig, coreutils
 Requires(preun): /sbin/chkconfig
+%{?systemd_requires}
 BuildRequires: glib2-devel popt-devel gettext pkgconfig systemd
 Provides: /sbin/service
 
@@ -57,21 +58,27 @@ touch %{buildroot}%{_sysconfdir}/rc.d/rc.local
 chmod 755 %{buildroot}%{_sysconfdir}/rc.d/rc.local
 
 %post
+%systemd_post fedora-import-state.service fedora-loadmodules.service fedora-readonly.service
+
 /usr/sbin/chkconfig --add network > /dev/null 2>&1 || :
 /usr/sbin/chkconfig --add netconsole > /dev/null 2>&1 || :
-if [ $1 -eq 1 ]; then
-  /usr/bin/systemctl daemon-reload > /dev/null 2>&1 || :
-fi
 
 %preun
+%systemd_preun fedora-import-state.service fedora-loadmodules.service fedora-readonly.service
+
 if [ $1 = 0 ]; then
   /usr/sbin/chkconfig --del network > /dev/null 2>&1 || :
   /usr/sbin/chkconfig --del netconsole > /dev/null 2>&1 || :
 fi
 
 %postun
-if [ $1 -ge 1 ]; then
-  /usr/bin/systemctl daemon-reload > /dev/null 2>&1 || :
+%systemd_postun fedora-import-state.service fedora-loadmodules.service fedora-readonly.service
+
+# This should be removed in Rawhide for Fedora 29:
+%triggerun -- initscripts < 9.78
+if [ $1 -gt 1 ]; then
+  systemctl enable fedora-import-state.service fedora-readonly.service &> /dev/null || :
+  echo -e "\nUPGRADE: Automatically re-enabling default systemd units: fedora-import-state.service fedora-readonly.service\n" || :
 fi
 
 %files -f %{name}.lang
--- a/systemd/system/fedora-import-state.service
+++ b/systemd/system/fedora-import-state.service
@@ -4,11 +4,14 @@ DefaultDependencies=no
 ConditionPathIsReadWrite=/
 ConditionDirectoryNotEmpty=/run/initramfs/state
 Conflicts=shutdown.target
-Before=shutdown.target emergency.service emergency.target systemd-tmpfiles-setup.service basic.target
+Before=shutdown.target emergency.service emergency.target systemd-tmpfiles-setup.service sysinit.target
 After=local-fs.target
 
 [Service]
-ExecStart=/lib/systemd/fedora-import-state
+ExecStart=/usr/lib/systemd/fedora-import-state
 Type=oneshot
 TimeoutSec=0
 RemainAfterExit=yes
+
+[Install]
+WantedBy=sysinit.target
--- a/systemd/system/fedora-loadmodules.service
+++ b/systemd/system/fedora-loadmodules.service
@@ -7,7 +7,10 @@ ConditionPathExists=|/etc/rc.modules
 ConditionDirectoryNotEmpty=|/etc/sysconfig/modules/
 
 [Service]
-ExecStart=/lib/systemd/fedora-loadmodules
+ExecStart=/usr/lib/systemd/fedora-loadmodules
 Type=oneshot
 TimeoutSec=0
 RemainAfterExit=yes
+
+[Install]
+WantedBy=sysinit.target
--- a/Makefile
+++ b/Makefile
@@ -64,15 +64,6 @@ install:
 	done
 
 # Can't store symlinks in a CVS archive
-	mkdir -p -m 755 $(ROOT)/usr/lib/systemd/system/multi-user.target.wants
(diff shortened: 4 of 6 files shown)


deekej:
We're still waiting for this BZ to be resolved first:
https://bugzilla.redhat.com/show_bug.cgi?id=1493479

deekej:
UPDATE: Pull-requst by @keszybz was merged into 'fedora-release` package.

deekej:
Thanks for spotting this, I wasnt't aware of this. I will update the commit. :)

deekej:
I have updated the specfile, but I'm unsure about few things:

* I have added the `Requires` of `systemd-units` for `post`, `postun` and `preun` phases.
* I have added `systemd-devel` instead of `systemd` (as suggested in the wiki)

I was looking at `httpd` package on how it does in Rawhide, and it is different compared to the wiki. But IMHO that shouldn't be a problem for the build/upgrade/removal of initscripts.

-----------------

I have also set the `fedora-import-state.service` and `fedora-readonly.service` to *not* be restarted during upgrade. These are `oneshot` type of services, and it doesn't make sense to me to run them during upgrade, it could break things. If I'm wrong, please correct me.

initscripts.spec
```diff
@@ -28,8 +28,11 @@ Conflicts: NetworkManager < 0.9.9.0-37.git20140131.el7
 Conflicts: ppp < 2.4.6-4
 Requires(pre): /usr/sbin/groupadd
 Requires(post): /sbin/chkconfig, coreutils
+Requires(post): systemd-units
+Requires(postun): systemd-units
 Requires(preun): /sbin/chkconfig
-BuildRequires: glib2-devel popt-devel gettext pkgconfig systemd
+Requires(preun): systemd-units
+BuildRequires: glib2-devel popt-devel gettext pkgconfig systemd-devel
```
keszybz: Please see the guidelines, they are up to date: https://fedoraproject.org/wiki/Packaging:Scriptlets#Systemd. `systemd-units` has been gone for many many years. You need systemd-devel only when linking to the systemd libraries.

initscripts.spec
```diff
@@ -57,19 +60,28 @@ touch %{buildroot}%{_sysconfdir}/rc.d/rc.local
 chmod 755 %{buildroot}%{_sysconfdir}/rc.d/rc.local
 
 %post
+%systemd_post fedora-import-state.service
+%systemd_post fedora-readonly.service
```
keszybz: It's better to put both services on the same line, to save systemctl invocations.

initscripts.spec
```diff
@@ -57,19 +60,28 @@ touch %{buildroot}%{_sysconfdir}/rc.d/rc.local
...
+
 /usr/sbin/chkconfig --add network > /dev/null 2>&1 || :
 /usr/sbin/chkconfig --add netconsole > /dev/null 2>&1 || :
 if [ $1 -eq 1 ]; then
   /usr/bin/systemctl daemon-reload > /dev/null 2>&1 || :
 fi
 
 %preun
+%systemd_preun fedora-import-state.service
+%systemd_preun fedora-readonly.service
```
keszybz: Here too.

initscripts.spec
```diff
@@ -57,19 +60,28 @@ touch %{buildroot}%{_sysconfdir}/rc.d/rc.local
...
+
 if [ $1 = 0 ]; then
   /usr/sbin/chkconfig --del network > /dev/null 2>&1 || :
   /usr/sbin/chkconfig --del netconsole > /dev/null 2>&1 || :
 fi
 
 %postun
+%systemd_postun fedora-import-state.service
+%systemd_postun fedora-readonly.service
```
keszybz: Here too.

systemd/system/fedora-import-state.service
```diff
@@ -12,3 +12,6 @@ ExecStart=/lib/systemd/fedora-import-state
 Type=oneshot
 TimeoutSec=0
 RemainAfterExit=yes
+
+[Install]
+WantedBy=local-fs.target
```
keszybz: I think this should be WantedBy=basic.target.

systemd/system/fedora-loadmodules.service
```diff
@@ -11,3 +11,6 @@ ExecStart=/lib/systemd/fedora-loadmodules
 Type=oneshot
 TimeoutSec=0
 RemainAfterExit=yes
+
+[Install]
+WantedBy=basic.target
```
keszybz: Hmm, most likely `WantedBy=sysinit.target`.

initscripts.spec
keszybz: And yes, using `%systemd_postun` (without restart) is correct.

deekej: OK, I will fix that. Just one more question - should I also add this line?
```
%{?systemd_requires}
```

I'm not sure about the purpose of that line...

systemd/system/fedora-import-state.service
deekej: This one will have to answer @lnykryn, I'm no expert on systemd. I have just used the targets as they were before (created by symlinks that we're removing).

systemd/system/fedora-loadmodules.service
deekej: Same here. @lnykryn?

systemd/system/fedora-import-state.service
lnykryn: I would put sysinit.target there, basic target seems a bit late to me.
@deekej just for FYI this is nicely documented in bootup manpage.

systemd/system/fedora-loadmodules.service
lnykryn: Yep we can do that.

systemd/system/fedora-import-state.service
deekej: After looking at the systemd bootup process details (and in context how it was done before and what it does), I agree with @lnykryn. The `basic.taget` can be already expecting that `fedora-import-state.service` has finished.

If you do not want the `local-fs.target` for that, then `sysinit.target` seems to be a way to go. :)

initscripts.spec
```diff
@@ -57,19 +58,31 @@ touch %{buildroot}%{_sysconfdir}/rc.d/rc.local
 chmod 755 %{buildroot}%{_sysconfdir}/rc.d/rc.local
 
 %post
+%systemd_post fedora-import-state.service \
+              fedora-loadmodules.service \
+              fedora-readonly.service
+
 /usr/sbin/chkconfig --add network > /dev/null 2>&1 || :
 /usr/sbin/chkconfig --add netconsole > /dev/null 2>&1 || :
 if [ $1 -eq 1 ]; then
```
lnykryn: I think that we can also drop the daemon-reload calls from the spec file. 
https://fedoraproject.org/wiki/Changes/systemd_file_triggers

initscripts.spec
keszybz: It generated the required Requires(pre) and Requires(post) lines.

systemd/system/fedora-import-state.service
keszybz: You're right, `sysinit.target` is better.

initscripts.spec
keszybz: No, please keep them as written in the guidelines. They evaluate to nothing if not necessary, but having the macro leaves the window open to change the implementation in systemd later on if necessary.

initscripts.spec
```diff
@@ -57,22 +58,32 @@ touch %{buildroot}%{_sysconfdir}/rc.d/rc.local
...
 %postun
-if [ $1 -ge 1 ]; then
-  /usr/bin/systemctl daemon-reload > /dev/null 2>&1 || :
-fi
+%systemd_postun fedora-import-state.service \
+                fedora-loadmodules.service \
+                fedora-readonly.service
+
+# This should be removed in Rawhide for Fedora 29:
+%triggerin -- initscripts > 9.77
```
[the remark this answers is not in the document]
deekej: Yes, we were discussing this with @lnykryn, because we were unsure about it. This will be changed.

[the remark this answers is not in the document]
deekej: I have updated this part. I need to make just one more test, but it seems promising.

keszybz:
LGTM.

deekej:
Unfortunately, we're still waiting on Dennis to make a new release of `fedora-release` package... :-/

PR #129, opened 2017-09-21
usleep: change the error message to print the full replacement commadline

keszybz:
Also drop the argv[0] as previx — the message already includes the
name of the program.

$ ./src/usleep 300
warning: usleep is deprecated, and will be removed in near future!
warning: use "sleep 0.3" instead...

%g will print a normal decimal point for all values that are likely to
be used (the smallest one possible is 1/1000 = 0.001). For very large
values scientific notation will be used, which is not very pretty, but
sleep accepts that, so there's no issue.

https://bugzilla.redhat.com/show_bug.cgi?id=1494168

### DIFF
--- a/src/usleep.c
+++ b/src/usleep.c
@@ -44,9 +44,6 @@ int main(int argc, char **argv) {
             { 0, 0, 0, 0, 0 }
         };
 
-  fprintf(stderr, "%s: warning: usleep(1) is deprecated, and will be removed in near future!!\n"
-                  "%s: warning: use sleep(1) instead...\n", argv[0], argv[0]);
-
   optCon = poptGetContext("usleep", argc, argv, options,0);
   /*poptReadDefaultConfig(optCon, 1);*/
   poptSetOtherOptionHelp(optCon, "[microseconds]");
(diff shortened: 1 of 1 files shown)


keszybz:
Oops, that's embarrassing. I pushed an updated version.

keszybz:
Pfff, it seems both %f and %g (even with very high precision) won't give the desired output (non-scientific notation without trailing zeros and without any rounding). So let's use %g, and just accept the fact that scientific notation will be used for very small values. It's help output anyway, so making something very complicated doesn't seem worth the trouble (and potential bugs).

Updated once again.

antofthy:
usleep() {
  # They bloody broke usleep by having it report an error
  # This ugly workaround (using scientific) seems to work...
  sleep ${1}e-06
}

kdudka:
`s/They/We/` because `usleep` is provided by initscripts itself.

antofthy:
And as it was available it got used by things other than initscripts, as the arguments are integers and thus easy to do basic math with in shell scripts.    Adding that little function just allows those scripts to work again, simply avoiding the need for floating point math in shell scripts using it!

PR #131, opened 2017-10-09
[ifup-post] Always update nameserver & search entries in /etc/resolv.conf

deekej:
This is complete rework of how we generate the `/etc/resolv.conf`.

#### Fixes:
 * always update `nameserver` & `search` entries when `DNS*` and `DOMAIN` entries (respectively) were updated in `ifcfg-*` files
 * always update the order of `nameserver` entries when the order of `DNS*` options was updated
 * always keep `/etc/resolv.conf.save` as a proper backup - do not overwrite it everytime when updating `nameserver` or `search`

#### Enhancements:
 * added support for `DNS3` option, since this is what *GLIBC* currently suppports
 * added logic to be able to process duplicate `DNS*` options
 * added logic to be able to process randomly omitted `DNS*` options
   (e.g. omitting `DNS1` while specifying `DNS2` and/or `DNS3` will still work now)

This work is based on these two RHEL-7.5 BZs: [RHBZ #1364895](https://bugzilla.redhat.com/show_bug.cgi?id=1364895) | [RHBZ #1357658](https://bugzilla.redhat.com/show_bug.cgi?id=1357658)

### DIFF
--- a/sysconfig/network-scripts/ifdown-post
+++ b/sysconfig/network-scripts/ifdown-post
@@ -18,35 +18,39 @@ source_config
 
 /etc/sysconfig/network-scripts/ifdown-routes ${REALDEVICE} ${DEVNAME}
 
-if [ "$PEERDNS" != "no" -o -n "$RESOLV_MODS" -a "$RESOLV_MODS" != "no" ] && \
+# Remove duplicate DNS entries and shift them,
+# to have always correct condition below...
+update_DNS_entries
+
+if ! is_false "${PEERDNS}" || ! is_false "${RESOLV_MODS}" && \
     [ "${DEVICETYPE}" = "ppp" -o "${DEVICETYPE}" = "ippp" -o -n "${DNS1}" \
     -o "${BOOTPROTO}" = "bootp" -o "${BOOTPROTO}" = "dhcp" ] ; then
-if [ -f /etc/resolv.conf.save ]; then
-    change_resolv_conf /etc/resolv.conf.save
-    rm -f /etc/resolv.conf.save
-fi
-if [ "${DEVICETYPE}" = "ppp" -o "${DEVICETYPE}" = "ippp" ]; then
-    if [ -f /etc/ppp/peers/$DEVICE ] ; then
-        rm -f /etc/ppp/peers/$DEVICE
+    if [ -f /etc/resolv.conf.save ]; then
+        change_resolv_conf /etc/resolv.conf.save
+        rm -f /etc/resolv.conf.save
     fi
-fi
+    if [ "${DEVICETYPE}" = "ppp" -o "${DEVICETYPE}" = "ippp" ]; then
+        if [ -f /etc/ppp/peers/$DEVICE ] ; then
+            rm -f /etc/ppp/peers/$DEVICE
+        fi
+    fi
 fi
 
 # Reset the default route if this interface had a special one
 if ! check_default_route ; then
     # ISDN device needs special handling dial on demand
     if [ "${DEVICETYPE}" = "ippp" -o "${DEVICETYPE}" = "isdn" ] && \
-        [ "$DIALMODE" = "auto" ] ; then
-    if [ -z "$GATEWAY" ] ; then
-        /sbin/ip route add default ${METRIC:+metric} \
-            ${WINDOW:+window $WINDOW} dev ${DEVICE}
+       [ "$DIALMODE" = "auto" ] ; then
+        if [ -z "$GATEWAY" ] ; then
+            /sbin/ip route add default ${METRIC:+metric} \
+                ${WINDOW:+window $WINDOW} dev ${DEVICE}
+        else
+            /sbin/ip route add default ${METRIC:+metric} \
+                ${WINDOW:+window $WINDOW} via ${GATEWAY}
+        fi
     else
-        /sbin/ip route add default ${METRIC:+metric} \
-            ${WINDOW:+window $WINDOW} via ${GATEWAY}
+        add_default_route ${DEVICE}
     fi
-else
-    add_default_route ${DEVICE}
-fi
 fi
 
 # Inform firewall
--- a/sysconfig/network-scripts/ifup-post
+++ b/sysconfig/network-scripts/ifup-post
@@ -28,67 +28,94 @@ if ! is_true "$NOROUTESET"; then
 fi
 
 
-if ! is_false "$PEERDNS" || [ -n "$RESOLV_MODS" ] && ! is_false "$RESOLV_MODS"; then
-    [ -n "$MS_DNS1" ] && DNS1=$MS_DNS1
-    [ -n "$MS_DNS2" ] && DNS2=$MS_DNS2
+if ! is_false "${PEERDNS}" || ! is_false "${RESOLV_MODS}"; then
+    # Obtain the DNS entries when using PPP if necessary:
+    [ -n "${MS_DNS1}" ] && DNS1="${MS_DNS1}"
+    [ -n "${MS_DNS2}" ] && DNS2="${MS_DNS2}"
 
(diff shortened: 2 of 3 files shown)


sysconfig/network-scripts/ifup-post
```diff
@@ -28,67 +28,111 @@ if ! is_true "$NOROUTESET"; then
...
+        while read line; do
+          case ${line} in
+
+            # Skip nameserver entries when at least one DNS option was given
+            # (at this stage we know that we have to update all the nameserver
+            # enries anyway -- see below), or copy them if we are changing just
+            # the 'search' field in /etc/resolv.conf:
+            nameserver*)
+              if [[ "${grep_regexp}" != ".*" ]]; then
+                continue
```
lnykryn: If I am not mistaken the original code only overwrote the first nameserver entries, but left the rest. THis one discards all of them. Also "${grep_regexp}" != ".*"  is really confusing.

sysconfig/network-scripts/ifup-post
```diff
@@ -28,67 +28,111 @@ if ! is_true "$NOROUTESET"; then
...
 
-        # backup resolv.conf
-        cp -af /etc/resolv.conf /etc/resolv.conf.save
+        # Add the requested nameserver entries:
+        [ -n "${DNS1}" ] && echo "nameserver ${DNS1}" >> "${tmp_file}"
+        [ -n "${DNS2}" ] && echo "nameserver ${DNS2}" >> "${tmp_file}"
+        [ -n "${DNS3}" ] && echo "nameserver ${DNS3}" >> "${tmp_file}"
+
+        # Backup resolv.conf only if it doesn't exist already:
+        ! [ -f /etc/resolv.conf.save ] && cp -af /etc/resolv.conf /etc/resolv.conf.save
```
lnykryn: Let's keep the old behavior here, for compatibility reasons.

sysconfig/network-scripts/ifup-post
```diff
@@ -28,67 +28,111 @@ if ! is_true "$NOROUTESET"; then
...
-        done
-        if [ -n "$DOMAIN" ]; then
-            echo "search $DOMAIN $search" >> $tr
+              echo "${line}" >> "${tmp_file}"
+              ;;
+          esac
+        done < /etc/resolv.conf
+
+        # Insert the domain into 'search' field:
+        if [[ -n "${DOMAIN}" ]]; then
```
lnykryn: Nitpick, [ ]  would be enough here and we use them everywhere in this case.

sysconfig/network-scripts/network-functions
```diff
@@ -448,6 +448,10 @@ set_hostname ()
             rsctmp=$(mktemp /tmp/XXXXXX);
             cat /etc/resolv.conf > $rsctmp
             echo "search $domain" >> $rsctmp
+
+            # Backup resolv.conf only if it doesn't exist already:
+            ! [ -f /etc/resolv.conf.save ] && cp -af /etc/resolv.conf /etc/resolv.conf.save
```
lnykryn: I don't think it is necessary to backup the resolv.conf in this case and I am bit worried what use-cases will this change break.

sysconfig/network-scripts/ifdown-post
```diff
@@ -18,7 +18,7 @@ source_config
 
 /etc/sysconfig/network-scripts/ifdown-routes ${REALDEVICE} ${DEVNAME}
 
-if [ "$PEERDNS" != "no" -o -n "$RESOLV_MODS" -a "$RESOLV_MODS" != "no" ] && \
+if ! is_false "${PEERDNS}" || ! is_false "${RESOLV_MODS}" && \
```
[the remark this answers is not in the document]
deekej: Oh, I haven't  realized that until you pointed that out... :-O That just proves the code needs indentation cleanup everywhere... :-/

sysconfig/network-scripts/ifdown-post
```diff
@@ -18,7 +18,7 @@ source_config
...
     [ "${DEVICETYPE}" = "ppp" -o "${DEVICETYPE}" = "ippp" -o -n "${DNS1}" \
```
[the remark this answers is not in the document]
deekej: Ah, good catch. I will fix that.

sysconfig/network-scripts/ifup-post
```diff
@@ -28,67 +28,111 @@ if ! is_true "$NOROUTESET"; then
...
+      grep_regexp="nameserver[[:space:]]+${DNS1}"
+    else
+      # No DNS entries used at all ->> match everything.
+      grep_regexp=".*"
+    fi
+
+    # Test if the search field needs updating, or
+    # if the nameserver entries order should be updated:
+    if [ -n "${DOMAIN}" ] && ! grep -q "^search.*${DOMAIN}.*$" /etc/resolv.conf ||
+       ! tr --delete '\n' < /etc/resolv.conf | grep -E -q "${grep_regexp}"; then
```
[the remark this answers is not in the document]
deekej: I'll look into the commented nameserver, as well the whitespaces.

However, I don't want to see the previous testing part there, because:
1) It's counter-intuitive. If user changes the order of DNS options (which is a valid thing to do), the nameserver entries WON'T be updated. This is clearly WRONG!
2) The problem of nameserve entries not updating is not even documented! And I would bet that NetworkManager properly updates the nameserver entries when the order of DNS options changes.
3) IMHO, this was the simplest solution (that I could came up with). If you find a better/simpler solution, I would like to hear about it. :) (No sarcasm intended.)

sysconfig/network-scripts/ifup-post
deekej: Yeah, this is just a leftover of previous bigger/more complicated condition...

sysconfig/network-scripts/ifup-post
deekej: I'm sorry, but no. What exactly is the old behaviour? In your words you described it as:

> Keep the backup of previous /etc/resolv.conf when user starts using network-scripts (network.service). And restore it, when they stop using the network-scripts.

The current behaviour does not comply to what you described, and it is such unpredictable and counter-intiutive behaviour of making `/etc/resolv.conf` backup that you can't be serious expecting anyone to rely on this behaviour...

If anyone would actually to understand how the logic for making `/etc/resolv.conf` backups works, they would have to hack the network-scripts to find out about it, most probably by trial and error approach - because this behaviour (AFAICT) is not even documented anywhere!

Therefore, if somebody was using this *undocumented* behaviour and expecting it to stay same forever, then it's their mistake to do so. This approach of hacking network-scripts (which is not sane approach) can't hold back fixes or progress. Period.

sysconfig/network-scripts/network-functions
deekej: I have actually spend more than a day by testing these (every corner-case I could think of) on RHEL-7.5 with these changes. This part is necessary for the changes to work as whole completely.

sysconfig/network-scripts/ifup-post
```diff
@@ -28,67 +28,111 @@ if ! is_true "$NOROUTESET"; then
...
-                if [ -n "$DOMAIN" ]; then
-                    echo "$answer" | while read key value ; do
-                        search="$search $value"
-                    done
-                else
-                    echo "$answer" >> $tr
-                fi
-                ;;
+    if [ -n "${DNS2}" ] && [[ "${DNS2}" == "${DNS1}" ]]; then
+      unset DNS2
```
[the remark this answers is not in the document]
deekej: Your comment is unclear, but I deduce that you mean this:
```
DNS1=A
DNS2=A
DNS3=C
```

In this case, the `DNS2` would end up unset, while the `DNS3` wouldn't be shifted. I have fixed that in my upcoming commit.

PR #132, opened 2017-10-10
[network-scripts] Replace firewall-cmd with DBus calls

deekej:
To speed things up, based on patch & comments from:
https://bugzilla.redhat.com/show_bug.cgi?id=1497759

### DIFF
--- a/sysconfig/network-scripts/ifdown-post
+++ b/sysconfig/network-scripts/ifdown-post
@@ -49,9 +49,12 @@ else
 fi
 fi
 
-# Inform firewall
-if [ -x /usr/bin/firewall-cmd -a "${REALDEVICE}" != "lo" ]; then
-    /usr/bin/firewall-cmd --remove-interface="${DEVICE}" > /dev/null 2>&1
+# Reset firewall ZONE to "default":
+if [ "${REALDEVICE}" != "lo" ]; then
+    dbus-send --system --dest=org.fedoraproject.FirewallD1 \
+              /org/fedoraproject/FirewallD1 \
+              org.fedoraproject.FirewallD1.zone.removeInterface \
+              string: "" string:"${DEVICE}"
 fi
 
 # Notify programs that have requested notification
--- a/sysconfig/network-scripts/ifup-eth
+++ b/sysconfig/network-scripts/ifup-eth
@@ -325,11 +325,6 @@ if [ -z "${NOZEROCONF}" -a "${ISALIAS}" = "no" -a "${REALDEVICE}" != "lo" ]; the
     ip route add 169.254.0.0/16 dev ${REALDEVICE} metric $((1000 + $(cat /sys/class/net/${REALDEVICE}/ifindex))) scope link
 fi
 
-# Inform firewall which network zone (empty means default) this interface belongs to
-if [ -x /usr/bin/firewall-cmd -a "${REALDEVICE}" != "lo" ]; then
-    /usr/bin/firewall-cmd --zone="${ZONE}" --change-interface="${DEVICE}" > /dev/null 2>&1
-fi
-
 if [ "${TYPE}" = "Bridge" ]; then
     for arg in $BRIDGING_OPTS ; do
         key=${arg%%=*};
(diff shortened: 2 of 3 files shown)


sysconfig/network-scripts/ifdown-post
```diff
@@ -49,9 +49,12 @@ else
...
 
-# Inform firewall
-if [ -x /usr/bin/firewall-cmd -a "${REALDEVICE}" != "lo" ]; then
-    /usr/bin/firewall-cmd --remove-interface="${DEVICE}" > /dev/null 2>&1
+# Reset firewall ZONE to "default":
+if [ "${REALDEVICE}" != "lo" ]; then
+    dbus-send --system --dest=org.fedoraproject.FirewallD1 \
+              /org/fedoraproject/FirewallD1 \
+              org.fedoraproject.FirewallD1.zone.removeInterface \
+              string: "" string:"${DEVICE}"
```
phoracek: just a nit: there is an empty space between string: and "", the other two calls don't have it.

sysconfig/network-scripts/ifdown-post
```diff
@@ -49,9 +49,12 @@ else
 fi
 fi
 
-# Inform firewall
-if [ -x /usr/bin/firewall-cmd -a "${REALDEVICE}" != "lo" ]; then
-    /usr/bin/firewall-cmd --remove-interface="${DEVICE}" > /dev/null 2>&1
+# Reset firewall ZONE to "default":
+if [ "${REALDEVICE}" != "lo" ]; then
+    dbus-send --system --dest=org.fedoraproject.FirewallD1 \
```
erig0: Unless you specify --print-reply this will be fire and forget. If you care about the result/exit code then you must specify --print-reply.

deekej: Yes, I realize that this will be *fire and forget*, and IMHO that was the point of the BZ in the first place -- they need a speedup for the network-scripts, and waiting for the reply & exit-code would not be so fast. (Correct me if I'm wrong.)

And lastly, we didn't check for the exit code before either, and after discussion with our colleagues we do not plan to check for it now.

sysconfig/network-scripts/ifdown-post
deekej: That's intentional -- please, see the related BZ's comments from @erig0 about why this is needed. ;)

phoracek: Gotcha, I though that only the "" matters.

sysconfig/network-scripts/ifdown-post
phoracek: In relative numbers, --print-reply adds quite a lot of time to the call. However, IMO compared to the original firewall-cmd it seems insignificant. I leave it to you if you think the error output would be worth it.

firewall-cmd --zone="${ZONE}" --change-interface="${DEVICE}"
> 0.309s

dbus-send --system --dest=org.fedoraproject.FirewallD1 /org/fedoraproject/FirewallD1 org.fedoraproject.FirewallD1.zone.changeZoneOfInterface string:"" string:"dummy_1"
> 0.005s

dbus-send --print-reply --system --dest=org.fedoraproject.FirewallD1 /org/fedoraproject/FirewallD1 org.fedoraproject.FirewallD1.zone.changeZoneOfInterface string:"" string:"dummy_1"
> 0.010s

deekej: We are actually not sure right now, what consequences the checking for exit codes / replies could have, especially in RHV environment with hundreds to thousands of interfaces. So far, we have been "successfully" avoiding the exit codes / replies, so in order to keep things working as they were before - especially in this case - we will keep silently ignoring the exit codes / replies.

erig0: Another thing worth pointing out is that this changes a synchronous call into an asynchronous call - it no longer blocks here. Any following commands or actions that depend on this command having already taken effect may be racey.

deekej: @erig0, that's actually a good point. There's one place where we might need to wait for the reply (i.e. synchronous call), and that's the IPv6 initialization. I will update the commit accordingly. :)

deekej:
I have updated the commint to include the *synchronous* call when obtaining IPv6 via DHCP:
https://github.com/fedora-sysv/initscripts/pull/132/commits/cf896bf9310a902912e0c6a4c4be5581ba8a1135#diff-1f650863fcc0e18c616793dca777ef28R341

The rest of the DBus calls can (IMHO) remain *asynchronous* for now.

sysconfig/network-scripts/ifdown-post
erig0: @phoracek, I think that was a typo on my part. I don't think the space is necessary.

PR #134, opened 2017-10-12
Drop sysconfig/init* files

deekej:
Users can still create the `/etc/sysconfig/init` file manually to override the default values, which are now taken from `/etc/rc.d/init.d/functions`.


deekej:
I have updated comments in the commit. Please review again, thanks. :)

PR #136, opened 2017-10-19
[specfile] drop dependency on /etc/system-release completely

deekej:
This way the `fedora-release` will be still installed by default together with initscripts, but initscripts will no longer block users from unistalling `fedora-release`, if they require so.

IOW, the Fedora branding will not be forced by initscripts any more.


deekej:
> Very dangerous, see https://bugzilla.redhat.com/show_bug.cgi?id=1493479#c19.

Agreed. I have reverted the change, and added the `27-1` version as discussed in the BZ. :)

deekej:
Based on https://bugzilla.redhat.com/show_bug.cgi?id=1493479#c35.

PR #137, opened 2017-10-19
[RHEL7] use weak (instead of strong) dependency for 'redhat-release'

deekej:
Same as #136, but for `redhat-release` and *RHEL-7*...

### DIFF
--- a/initscripts.spec
+++ b/initscripts.spec
@@ -22,8 +22,7 @@ Conflicts: lvm2 < 2.02.100-5
 Conflicts: dmraid < 1.0.0.rc16-18
 Requires: systemd
 Requires: iproute, /sbin/arping, findutils
-# Not strictly required, but nothing else requires it
-Requires: /etc/system-release
+Recommends: redhat-release
 Requires: udev >= 125-1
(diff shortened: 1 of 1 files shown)


initscripts.spec
```diff
@@ -22,8 +22,7 @@ Conflicts: lvm2 < 2.02.100-5
 Conflicts: dmraid < 1.0.0.rc16-18
 Requires: systemd
 Requires: iproute, /sbin/arping, findutils
-# Not strictly required, but nothing else requires it
-Requires: /etc/system-release
+Recommends: redhat-release
```
lnykryn: We don't have weak dependencies in EL7 and please check that redhat-release packages actually exists.

deekej: Yes, I intend to change it to this:

```
Conflicts: redhat-release < <version>
```

and removing the `Recommends`. We still need to wait for the new version of `redhat-release`, so we can put the correct version to the conflicts.

deekej:
I have changed my mind after thinking about this. I'm closing this PR and creating a new one instead.

PR #142, opened 2017-10-23
Avoid some unnecessary stat calls

scop:


deekej:
Hello @scop, thank you for all the pull-requests. :)

However, for this one, I will need some deeper description. Could you please elaborate a little why we should incorporate these changes? At some hunks I'm not sure what the benefit is. Thank you. ;)

rc.d/init.d/functions
```diff
@@ -59,7 +59,7 @@ systemctl_redirect () {
 [ -z "${COLUMNS:-}" ] && COLUMNS=80
 
 if [ -z "${CONSOLETYPE:-}" ]; then
-    if [ -c "/dev/stderr" -a -r "/dev/stderr" ]; then
+    if [ -c "/dev/stderr" ] && [ -r "/dev/stderr" ]; then
```
deekej: I'm really not sure about the benefit of this - could you please explain? :)

deekej:
OK, for clarification: https://github.com/koalaman/shellcheck/wiki/SC2166

scop:
To clarify, the primary motivation of this is not the well-definedness (which was kind of new to me, thanks for the pointer), but rather short-circuiting; `-o` and `-a` do not short circuit, meaning both sides will always be evaluated, resulting in unnecessary stat calls in this particular case. Splitting to two statements separated by `&&` or `||` provides the desired short-circuit behavior.

deekej:
Yes, you're right. :) This was another thing we have find out with our colleagues while discussing your pull-request... ;) It's also better readable, so I hope to fix the code to use it everywhere one day. Thanks for it! :)

PR #154, opened 2017-11-27
network: add knob to optionally keep interfaces up during shutdown

davide125:
On machines that use systemd as PID1 but don't enable pam_systemd, SSH sessions aren't terminated properly on shutdown (see https://bugzilla.redhat.com/show_bug.cgi?id=1047614 for details). This PR adds a knob to workaround the issue, by optionally leaving the network up during shutdown. This is only implemented for systemd hosts, but it could be extended to SysV ones as well if desired.

### DIFF
--- a/rc.d/init.d/network
+++ b/rc.d/init.d/network
@@ -167,6 +167,15 @@ stop)
         exit 1
     fi
 
+    # Don't shut the network down when shutting down the system if configured
+    # as such in sysconfig
+    if [ "$IFDOWN_ON_SHUTDOWN" = no ]; then
+      if systemctl is-system-running | grep -q 'stopping'; then
+        net_log "system is shutting down, leaving interfaces up as requested"
+        exit 1
+      fi
+    fi
+
     vlaninterfaces=""
     vpninterfaces=""
     xdslinterfaces=""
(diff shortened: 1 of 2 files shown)


deekej:
Hello guys! Long time no see. :) Thanks for the pull-request, it seems reasonable to us to accept it. Just a few minor things...

rc.d/init.d/network
```diff
@@ -167,6 +167,15 @@ stop)
         exit 1
     fi
 
+    # Don't shut the network down when shutting down the system if configured
+    # as such in sysconfig
+    if [ "$IFDOWN_ON_SHUTDOWN" = no ]; then
```
deekej: Please, from now on, use `is_true` or `is_false` functions for testing of options' values, since these functions also process other values and we try to use it everywhere now... :) See here:
https://github.com/fedora-sysv/initscripts/blob/master/rc.d/init.d/functions#L654

rc.d/init.d/network
```diff
@@ -167,6 +167,15 @@ stop)
...
+      if systemctl is-system-running | grep -q 'stopping'; then
+        net_log "system is shutting down, leaving interfaces up as requested"
```
deekej: One more small thing - please use the `net_log $"log/warning/error message"` format when using the `net_log` function (note the additional `$`). This allows these logging messages to be translated by volunteers into other languages. (String starting with `$` are marked for translations.) Thank you! :)

davide125:
Thanks @deekej, I've updated this to address your review.

PR #162, opened 2018-01-30
service: check for 'status' support when doing '--status-all'

deekej:
To avoid calling of `status` on some non-LSB compliant initscripts. This is not perfect and is *best-effort only*, because some crazily written initscripts might have different way of checking/displaying of `status`.

We print a warning when we stumble on a script that we assume does not have the `status` functionality implemented. I expect us to receive a bug report for a script that implements `status` in some strange way, so we will add the regex pattern for it if needed.


lnykryn:
Ugh, definitely no. Initscripts have a defined interface that they must follow. Also, this "detection" would break in so many cases...

PR #170, opened 2018-03-12
specfile: remove the systemctl call in the %postun phase

deekej:
See https://bugzilla.redhat.com/show_bug.cgi?id=1512688 for more info.

NOTE: We shouldn't merge this commit until F28 is released as GA - to make sure the upgrades go smoothly... :)


deekej:
Fedora 28 is live now, merging...

PR #171, opened 2018-03-13
network.service: 'reload' removed

deekej:
More info: https://bugzilla.redhat.com/show_bug.cgi?id=1554690


deekej:
Yes, you're right. I'll update the commit.

PR #172, opened 2018-03-23
ifup-tunnel: Support 'external' tunnels

jaymzh:
This is a newish feature upstream. You can now set the external flag on a
ip6_tunnel type interface (though not the primary one, ip6tnl0), and doing so
will allow it to decapsulate any packet, and assuming that the inner address is
the one on that interface, it'll drop it back on the stack.

This is useful for DSR vips. While v6-in-v6 was already supported, this allows
v4-in-v6 which is necessary to serve v4 traffic in a v6only infrastructure.

There's comments in ifup-tunnel that imply it was designed only for GRE tunnels,
but this still seems like the best place for this.

### DIFF
--- a/sysconfig.txt
+++ b/sysconfig.txt
@@ -931,7 +931,12 @@ Files in /etc/sysconfig/network-scripts/
       "mode=active-backup arp_interval=60 arp_ip_target=192.168.1.1,192.168.1.2"
 
   Tunnel-specific items:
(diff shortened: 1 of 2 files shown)


jaymzh:
I've added docs. Happy to add more if you want, just ping me via email or assign an Issue to me. Gonna merge for now.

PR #174, opened 2018-04-09
use proper shebang

teknoraver:
$" is a BASH extension to translate text.
Replace the shebang with /bin/bash to avoid parsing errors,
as even BASH has a POSIX compliant mode when run as sh.

### DIFF
--- a/service
+++ b/service
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 . /etc/init.d/functions
 
--- a/sys-unconfig
+++ b/sys-unconfig
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 . /etc/init.d/functions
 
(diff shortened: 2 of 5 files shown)


deekej:
Hmm, I like the idea, but... IIRC, there are some people that use the initscrits/network-scripts with non-BASH shells.

First I need to check if this won't break something for them... :)

teknoraver:
I really hope that some people are using other shells as default shell (I'm a dash fan), as BASH is very slow in processing scripts.
Altough if they do, scripts with bashism won't work. In this case fortunately, it will just print a spurious dollar sign.
```
$ bash -c 'echo $"Hello" '
Hello
$ dash -c 'echo $"Hello" '
$Hello
$ zsh -c 'echo $"Hello" '
$Hello
```

deekej:
OK, this looks good to me. I will merge it into the next release... :)

deekej:
I have merged this into master branch for Fedora. However, I'm not sure this should make it into RHEL-7 branch. There's still chance it could break something for our customers...

PR #175, opened 2018-04-09
network: add option to keep the link down

teknoraver:
Some interfaces like Open vSwitch bridges using the userspace datapath,
needs the link to be kept down to avoid issues.
Add a LINKSTATUS=[down|up] ifcfg parameter to add this functionality.
If not specified, the link will default to up as before.

### DIFF
--- a/sysconfig/network-scripts/ifup
+++ b/sysconfig/network-scripts/ifup
@@ -116,7 +116,7 @@ if [ "${VLAN}" = "yes" ] && [ "$ISALIAS" = "no" ] && [ -n "$DEVICE" ]; then
         }
 
         # Link on Physical device needs to be up but no ip required
-        check_device_down ${PHYSDEV} && { ip -o link set dev ${PHYSDEV} up; }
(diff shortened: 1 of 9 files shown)


lnykryn:
Maybe a stupid question, but anyway: How does this work with Network Manager?

deekej:
> Maybe a stupid question, but anyway: How does this work with Network Manager?

That's a good question, actually... :) Will the NetworkManager do the same as well when it sees `LINKSTATUS=down`? Or will we introduce a different behaviour from NetworkManager?

deekej:
>  NetworkManager will ignore LINKSTATUS=down and bring up the interface anyway.

OK, this is something I'd like to avoid. Introducing new feature that is not yet in NetworkManager is a bad thing to do nowadays. We need guys from NetworkManager to look at this as well... I'll talk to them.

teknoraver:
do we have a bug tracked for thecorresponding NM feature request?

deekej:
> do we have a bug tracked for the corresponding NM feature request?

Not that I'm aware of... @lkundrak might know better, but I think we might have to create it first.

Issue #179, opened 2018-04-27
logic of is_hostname_set is incorrect


LipingMao:
Hi ,

The logic of is_hostname_set in not right, it introduced in [1]:
is_hostname_set ()
{
    case "${HOSTNAME}" in
        '(none)' | 'localhost' | 'localhost.localdomain')
            # Hostname NOT set:
            return 1
            ;;
        *)
            # Hostname IS set:
            return 0
            ;;
    esac
}

when Hostname NOT set, it should return 0, When Hostname IS set, it should return 1.
Otherwise, need_hostname will return a wrong value in the case:
NO_DHCP_HOSTNAME is not set, and hostname is already set.
If dhcpclient request hostname option, it will overwrite existed hostname.
Thanks.

[1] https://github.com/fedora-sysv/initscripts/commit/2ea1dcf2d22606f4b30ae16a14a736d0d9dddf24#diff-4003056111e3c781065006ea15b573ea

deekej:
Hello,

I think you're missing the important fact that shell has inverted truth/false values. In shell, return value of 0 means success, and other different value means some failure. You need to look at this as the exit values of the each program/command, which can behave in the same way.

I acknowledge it can be sometimes quite confusing, but for now I don't see a reason to merge this pull-request.

Best regards,

Dee'Kej

LipingMao:
@deekej  Thanks for the reply, I was noticed that after 4 days (Orz), so I closed it.  ;-)

PR #180, opened 2018-04-27
fix is_hostname_set wrong return value

LipingMao:
is_hostname_set should return 1 when hostname is set, return 0 when hostname is not set.


deekej:
Please, see the comment https://github.com/fedora-sysv/initscripts/issues/179#issuecomment-385932916.

PR #183, opened 2018-05-02
network-functions: add error messages for bonding installation

deekej:
Instead of displaying messages without context, like this:

```
./network-functions: line 571: echo: write error: Invalid argument
./network-functions: line 598: echo: write error: Permission denied
```

We will now display pretty error messages (via `net_log`) about what has actually failed...

### DIFF
--- a/sysconfig/network-scripts/network-functions
+++ b/sysconfig/network-scripts/network-functions
@@ -554,6 +554,8 @@ bond_master_exists ()
 
 install_bonding_driver ()
 {
+    local fn="install_bonding_driver"
+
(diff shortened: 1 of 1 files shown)


sysconfig/network-scripts/network-functions
```diff
@@ -554,6 +554,8 @@ bond_master_exists ()
 
 install_bonding_driver ()
 {
+    local fn="install_bonding_driver"
```
lnykryn: In the past, the last argument of net_log function was only used in ipv6 part of the code. But meh.

PR #184, opened 2018-05-02
RHEL-7 | network-functions: add error messages for bonding installation

deekej:
Same as #183. This has been requested by our customer:
https://bugzilla.redhat.com/show_bug.cgi?id=1542514

### DIFF
--- a/sysconfig/network-scripts/network-functions
+++ b/sysconfig/network-scripts/network-functions
@@ -550,6 +550,8 @@ bond_master_exists ()
 
 install_bonding_driver ()
 {
+    local fn="install_bonding_driver"
+
     if ! bond_master_exists ${1}; then
         modprobe bonding || return 1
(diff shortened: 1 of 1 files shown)


sysconfig/network-scripts/network-functions
```diff
@@ -550,6 +550,8 @@ bond_master_exists ()
 
 install_bonding_driver ()
 {
+    local fn="install_bonding_driver"
```
lnykryn: In the past, the last argument of net_log function was only used in ipv6 part of the code. But meh.

deekej: Yeah, I think we should unite this behaviour for all the functions. I hope it might ease up debugging in the future, especially when customer would send initially the contents of the `journalctl -b` (without the `rc.debug` on the kernel command line).

PR #190, opened 2018-05-10
RHEL-7 | sysconfig/readonly-root: Clarify the usage of readonly-root.

deekej:
Same problem as in #189. More info: https://bugzilla.redhat.com/show_bug.cgi?id=1444018


sysconfig/readonly-root
```diff
@@ -1,4 +1,5 @@
 # Set to 'yes' to mount the system filesystems read-only.
+# NOTE: It's necessary to append 'ro' to mount options of '/' mount point as well.
```
lnykryn: Please mention that it needs to be done in /etc/fstab, so we avoid the confusion with ro and rw on kernel cmdline.

PR #195, opened 2018-05-14
RHEL-7 | rhel-autorelabel: set UEFI BootNext == BootCurrent

deekej:
In some circumstances the system could be started into UEFI shell instead of RHEL after the rhel-autorelabel has run... This commit should fix this behaviour.

### DIFF
--- a/systemd/rhel-autorelabel
+++ b/systemd/rhel-autorelabel
@@ -5,6 +5,30 @@
 
 . /etc/init.d/functions
 
+# If the user has this (or similar) UEFI boot order:
+#
+#             Windows | grub | Linux
+#
+# And decides to boot into grub/Linux, then the reboot at the end of autorelabel
+# would cause the system to boot into Windows again, if the autorelabel was run.
+#
+# This function restores the UEFI boot order, so the user will boot into the
+# previously set (and expected) partition.
+efi_set_boot_next() {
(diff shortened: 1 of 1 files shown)


lnykryn:
I have no idea here, the proposed change kinda makes sense, but I am a bit worried about existing setups.

deekej:
This patch should make it first into `selinux` package, so we have it in upstream first.

deekej:
Pull-request submitted to upstream: https://src.fedoraproject.org/rpms/policycoreutils/pull-request/5

systemd/rhel-autorelabel
```diff
@@ -31,6 +31,17 @@ relabel_selinux() {
         [ -x "/usr/sbin/quotaoff" ] && /usr/sbin/quotaoff -aug
 	/sbin/fixfiles $FORCE restore > /dev/null 2>&1
     fi
+
+    # Check if the efibootmgr is installed and we are on EFI-enabled system:
+    # (NOTE: On non-EFI-enabled system the efibootmgr will fail.)
+    if efibootmgr > /dev/null 2>&1; then
+        # Restore the BootNext value unless it is already set (by something):
```
msekletar: I'd think about introducing a function e.g. ```efi_set_boot_next()``` and providing a rationale (as a comment) why is function needed. I'd drop the comments that explain what code *does* since it is obvious.

deekej:
The upstream pull-request has been merged.

PR #196, opened 2018-05-17
RHEL-7 | network-functions: use tr to upper case strings rather than awk

deekej:
Backport of PR #176.

--------------------

network-functions uses awk to tranform the mac address to upper case.
Replace awk with tr which is two order of magnitude faster:

    $ time awk '{ print toupper($0) }' </usr/share/dict/words >/dev/null

    real    0m0,227s
    user    0m0,224s
    sys     0m0,004s
    $ time tr '[a-z]' '[A-Z]' </usr/share/dict/words >/dev/null

    real    0m0,005s
    user    0m0,000s
    sys     0m0,005s

Also use here-strings instead of spawning a subshell and a pipe.

### DIFF
--- a/sysconfig/network-scripts/network-functions
+++ b/sysconfig/network-scripts/network-functions
@@ -16,7 +16,7 @@ HOSTNAME="$(hostname)"
 get_hwaddr ()
 {
     if [ -f /sys/class/net/${1}/address ]; then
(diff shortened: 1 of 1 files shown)


AsciiWolf:
lgtm!

PR #198, opened 2018-05-24
Spring Cleanup

deekej:
This is the first part of the "spring cleanup" process, mostly moving or removing of files that should no longer be part of initscripts. Please, follow up every commit to see the files moved / removed.


lang.csh
```diff
@@ -1,83 +1,78 @@
-# /etc/profile.d/lang.csh - set i18n stuff
+# /etc/profile.d/lang.csh - exports environment variables, and provides fallback
+#                          for CJK languages that can't be displayed in console.
```
keszybz: Indentation.

lang.sh
```diff
@@ -1,90 +1,65 @@
...
-else
-    for langfile in /etc/locale.conf "$HOME/.i18n" ; do
-        [ -f $langfile ] && . $langfile && sourced=1
-    done
+if [ -n "${LANG_backup}" ]; then
+    LANG="${LANG_backup}"
 fi
 
-if [ "$sourced" = 1 ]; then
-    [ -n "$LANG" ] && export LANG || unset LANG
```
keszybz: You remove this part, i.e. the unsetting of variables that have empty values. In particular, the `sed` expression above does not allow `unset` if I'm reading the regexp correctly, so setting the variable to an empty value seems to have been the way to get rid of a setting. So I think this "unsetting" part should be kept.

lang.sh
```diff
@@ -1,90 +1,65 @@
...
+    if grep --quiet -E -i -e '^.+\.utf-?8$' <<< "${LANG}"; then
+        case ${LANG} in
+            ja*)    LANG=en_US.UTF-8 ;;
+            ko*)    LANG=en_US.UTF-8 ;;
+            si*)    LANG=en_US.UTF-8 ;;
+            zh*)    LANG=en_US.UTF-8 ;;
+            ar*)    LANG=en_US.UTF-8 ;;
+            fa*)    LANG=en_US.UTF-8 ;;
+            he*)    LANG=en_US.UTF-8 ;;
+            en_IN*) true             ;;
```
keszybz: `something) ;;` is idiomatic in shell scripts, calling `true` here seems strange.

lang.csh
keszybz: If's off by one, the "f" should be under "e".

keszybz: Well "should" for some definition of "should" ;)
At least in the other file it's aligned.

deekej: Ah, I see now. :) Thanks, I'll fix it :)

deekej:
I've included a commit (with modifications so it correctly applies to new repository layout) from @lkundrak, to use the alternatives system as well. This resolves #197 also.

PR #202, opened 2018-05-29
Repository & specfile cleanup

deekej:
This PR establishes new repository layout for easier orientation & maintenance inside it. As consequences, the Makefiles and specfile are cleaned & update as well.

This change should make additional planned changes for the initscripts package more convenient/easier to implement.

NOTE: It's easier to follow up each commit and its changes, rather than just viewing the change in PR as a whole.


deekej:
I've put the symlink there just out of convenience, but we can remove it in the future if it'll lead to some issues. :)

PR #205, opened 2018-05-31
Spring cleanup #2

deekej:
Another round of fixes & modifications.


deekej:
Bumspec added. It is ready to be released in Rawhide once we're synced with other packages.

PR #207, opened 2018-06-04
specfile: (sub)packages split

deekej:
From now on, we will have 4 separate packages (builded from one SRPM):
 * initscripts
 * network-scripts
 * netconsole-service
 * readonly-root

`network-scripts`, `netconsole-service`, and `readonly-root` all require the `initscripts` package (as a parent).

NOTE: The translations for the `network-scripts` package remained in `initscripts`, because some of those translations are also needed in `initscripts`, and splitting those translations is too much PITA.


initscripts.spec
```diff
@@ -124,6 +121,45 @@ allows logging of kernel messages over the network.
...
+%description -n network-scripts
+This package contains the legacy scripts for activating & deactivating of most
+network interfaces. It also provides a legacy version of 'network' service.
+
+The 'network' service is enabled by default after installation of this package,
+and if the network-scripts are installed alongside NetworkManager, then the
+ifup/ifdown commands from network-scripts take precedence over the ones provided
+by NetworkManager.
+
+If uses has both netwokrk-scripts & NetworkManager installed, and wishes to
```
AsciiWolf: Small typos:
uses -> user
netwokrk-scripts -> network-scripts
they has -> they have

initscripts.spec
```diff
@@ -100,6 +86,87 @@ Conflicts:        systemd            < 216-3
...
+Requires:         iproute
+Requires:         ipcalc
+Requires:         util-linux
+
+%description -n readonly-root
+This package provides script & configuration file for setting up read-only root
+support. However, the script is quite buggy and is limited in what it can do.
+
+It is suggested to not use this package, unless really necessary. Additional
+configuration is required after installation.
```
lnykryn: I don't think that subjective evaluation belongs here. We have a lot of users that are happily using this and more importantly, we don't have any replacement.

initscripts.spec
deekej: Thanks, I'll fix it. :)

initscripts.spec
```diff
@@ -206,24 +277,59 @@ fi
...
+
+%config(noreplace)    %{_sysconfdir}/sysconfig/network-scripts/ifcfg-lo
+
+%ghost                %{_sbindir}/ifup
+%ghost                %{_sbindir}/ifdown
+%attr(4755,root,root) %{_sbindir}/usernetctl
+
+%{_mandir}/man8/ifup.*
+%{_mandir}/man8/ifdown.*
+%{_mandir}/man8/usernetctl.*
```
lnykryn: /usr/share/doc/initscripts/examples/networking/* should be here as well

lnykryn: Maybe /usr/sbin/usernetctl should be provided by NM as well? Dunno

initscripts.spec
deekej: I forgot to add the `initscripts-compat` package for this.

deekej:
> How do we handle upgrades? If we put this in Fedora users will have half of the tools uninstalled during the upgrade.

I forgot to add the `initscripts-compat` package for this.

initscripts.spec
deekej: It actually already is: https://github.com/fedora-sysv/initscripts/pull/207/files#diff-19805c24876f7b40ffecaa684780ae74R307

deekej: I'm on it.

initscripts.spec
```diff
@@ -99,6 +99,21 @@ network interfaces, some utilities, and other legacy files.
 
 # === SUBPACKAGES =============================================================
 
+%package compat
+Summary:          Compatibility package for upgrade purposes
+
+Obsoletes:        %{name}            < 9.82-2
+Requires:         netconsole-service = %{version}-%{release}
+Requires:         network-scripts    = %{version}-%{release}
+Requires:         readonly-root      = %{version}-%{release}
```
keszybz: You should not need a compat package. https://fedoraproject.org/wiki/Upgrade_paths_%E2%80%94_renaming_or_splitting_packages#.28n:1.29_Many-to-one_replacement describes the scenario you need (n:1 replacement). But also see https://bugzilla.redhat.com/show_bug.cgi?id=1261034. I *think* dnf now does the upgrades properly, but this should be tested.

deekej: I actually tested this. If you upgrade the new initscripts package (without existing `compat` subpackage), then the `network-scripts`, `readonly-root`, and `netconsole-service` won't be installed. If this happens, it could break some existing setups - e.g. the systems without NetworkManager might not connect to network after boot up. This might be dangerous for some virtual machines, etc.

initscripts.spec
```diff
@@ -189,6 +210,17 @@ if [ $1 -eq 0 ]; then
...
+# ---------------
+
+%post -n readonly-root
+%systemd_post readonly-root.service
+
+%preun -n readonly-root
+%systemd_post readonly-root.service
+
+%postun -n readonly-root
+%systemd_post readonly-root.service
```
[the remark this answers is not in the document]
deekej: Good catch - copy & paster error, gonna fix it. :)

initscripts.spec
```diff
@@ -189,6 +210,17 @@ if [ $1 -eq 0 ]; then
   chkconfig --del netconsole > /dev/null 2>&1 || :
 fi
 
+# ---------------
+
+%post -n readonly-root
+%systemd_post readonly-root.service
+
+%preun -n readonly-root
+%systemd_post readonly-root.service
```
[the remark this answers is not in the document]
deekej: Good catch - copy & paster error, gonna fix it. :)

initscripts.spec
```diff
@@ -100,6 +100,28 @@ Conflicts:        systemd            < 216-3
...
+Requires:         iproute
+Requires:         iputils
+Requires:         kmod
+Requires:         sed
+Requires:         util-linux
+
+%description -n netconsole-service
+This packages provides a 'netconsole' service for loading of netconsole kernel
+module with the configured parameters. The netconsole kernel module itself then
+allows logging of kernel messages over the network.
```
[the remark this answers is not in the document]
deekej: > Pfff, maybe instead of repackaging the existing sysvinit script, the script could be turned into a proper systemd .service unit.

I'm generally OK with this. However, right now I need to the package split of network-scripts ASAP, since I'm going on vacation soon, and we need this change for the next major version of RHEL.

I'll talk with Lukas about this. Thanks! :)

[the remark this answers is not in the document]
lnykryn: We have agreed that we will wrap this in a proper unit file in following pull request. But this will be done before next upstream release.

initscripts.spec
```diff
@@ -124,6 +121,44 @@ allows logging of kernel messages over the network.
 
 # ---------------
 
+%package -n network-scripts
```
lnykryn: since the network-scripts package actually ships an init scripts it should have a dependency on chkconfig (preun,post)

initscripts.spec
```diff
@@ -122,19 +202,22 @@ network interfaces, some utilities, and other legacy files.
...
 touch %{buildroot}%{_sbindir}/ifdown
 
-# ---------------
+# =============================================================================
 
 %post
-%systemd_post import-state.service loadmodules.service readonly-root.service
+%systemd_post import-state.service loadmodules.service
 
 chkconfig --add network > /dev/null 2>&1 || :
```
lnykryn: Do we actually need this chkconfig call? Those initscripts are not enabled in the chkconfig header, so maybe it is enough just to leave it systemd macros? But we still need the dependency on chkconfig, since the systemd might call it in the end.

lnykryn: Scratch that, we need those.

lnykryn: But they need to be in network-scrips subpackage

keszybz:
Please post the version without compat package (can be a separate branch if you prefer…). That is the documented way that such splits are supposed to be done, so it is expected to work. If it is really broken, I want to file bugs against dnf.

deekej:
This is not about broken DNF, it's about how we do things in this change. DNF by default does not keep packages that were not installed by the user, and which are no longer needed as dependencies. Which is correct behaviour.

In this case, users of F28 (e.g.) have only `initscripts` package installed by default, and we are moving some of the stuff from initscripts into different packages - namely `network-scripts`.

That means that during the upgrade, only new initscripts package will be installed. This package does not contain the network-scripts anymore, and DNF does not have a reason to install network-scripts package. When this happens, users will end up **without** network-scripts on their system, which can ultimately lead to broken systems. (People might not be able to connect to virtual machines any more, if they had only initscripts package installed before.)

I have tested this behaviour in Fedora Rawhide, and without the *-compat subpackage the things will end up broken...

deekej:
Here's the branch without *-compat subpackage:
https://github.com/deekej/initscripts/tree/enhance/package-split-without-compat

keszybz:
The guideline page and the bug I linked are all about what you are saying **not** being true.
A quick look at `git diff origin/master..deekej/enhance/package-split-without-compat *spec` shows that "Obsoletes" does not appear in the diff *at all*. Please implement the change as described by the guidelines, and if *that* doesn't work, let's figure out why.

deekej:
`*-compat` subpackage has been dropped.

PR #208, opened 2018-06-05
l10n: add zanata.xml

AsciiWolf:


deekej:
Do we have any idea how to automatically sync Zanata & initscripts repository?

Also, if we use Zanata, does it mean that `.tx/` can be removed?

lnykryn:
Yep we can drop the .tx directory.

deekej:
@AsciiWolf, please create another commit which drops the `.tx/` directory and I'll will merge the changes... ;)

PR #212, opened 2018-06-11
network-scripts: Replace brctl with ip-link

n0-1:
Since ip-link has full support for Linux bridges (and slave ports), use
that instead of the deprecated brctl from bridge-utils.

Beware: These changes are completely untested! Thorough review/testing is therefore strongly advised. Also this completely ignores existing documentation, which probably needs updating to cover the changes. Though backwards compatibility has been considered when writing the new code.

### DIFF
--- a/network-scripts/ifdown-eth
+++ b/network-scripts/ifdown-eth
@@ -135,14 +135,13 @@ if [ -d "/sys/class/net/${REALDEVICE}" ]; then
 fi
 [ "$retcode" = "0" ] && retcode=$?
 
-if [ -n "${BRIDGE}" ] && [ -x /usr/sbin/brctl ]; then
-    /sbin/ip link set dev ${DEVICE} down
-    /usr/sbin/brctl delif -- ${BRIDGE} ${DEVICE}
+if [ -n "${BRIDGE}" ]; then
+    ip link set dev ${DEVICE} nomaster down
     # Upon removing a device from a bridge,
     # it's necessary to make radvd reload its config
     [ -r /run/radvd/radvd.pid ] && kill -HUP $(cat /run/radvd/radvd.pid)
     if [ -d /sys/class/net/${BRIDGE}/brif ] && [ $(ls -1 /sys/class/net/${BRIDGE}/brif | wc -l) -eq 0 ]; then
-        /usr/sbin/brctl delbr -- ${BRIDGE}
+        ip link del ${BRIDGE}
     fi
 fi
 
--- a/network-scripts/ifup-eth
+++ b/network-scripts/ifup-eth
@@ -45,21 +45,24 @@ if [ -n "${HWADDR}" ]; then
     fi
 fi
 
-# If the device is a bridge, create it with brctl, if available.
+# If the device is a bridge, create it
 if [ "${TYPE}" = "Bridge" ]; then
-    if [ ! -x /usr/sbin/brctl ]; then
-        net_log $"Bridge support not available: brctl not found"
-        exit 1
+    bridge_opts=""
+    [ -n "${DELAY}" ] && bridge_opts+="forward_delay ${DELAY} "
+    if is_true "${STP}"; then
+	    bridge_opts+="stp_state 1 "
+    elif is_false "${STP}"; then
+	    bridge_opts+="stp_state 0 "
     fi
+    [ -n "${PRIO}" ] && bridge_opts+="priority ${PRIO} "
+    [ -n "${AGEING}" ] && bridge_opts+="ageing_time ${AGEING} "
 
     if [ ! -d /sys/class/net/${DEVICE}/bridge ]; then
-        /usr/sbin/brctl addbr -- ${DEVICE} || exit 1
+        ip link add ${DEVICE} type bridge $bridge_opts
+    elif [ -n "${OPTS}" ]; then
+        ip link set ${DEVICE} type bridge $bridge_opts
     fi
-
-    [ -n "${DELAY}" ] && /usr/sbin/brctl setfd -- ${DEVICE} ${DELAY}
-    [ -n "${STP}" ] && /usr/sbin/brctl stp -- ${DEVICE} ${STP}
-    [ -n "${PRIO}" ] && /usr/sbin/brctl setbridgeprio ${DEVICE} ${PRIO}
-    [ -n "${AGEING}" ] && /usr/sbin/brctl setageing ${DEVICE} ${AGEING}
+    unset bridge_opts
 
     # add the bits to setup driver parameters here
     for arg in $BRIDGING_OPTS ; do
@@ -173,15 +176,15 @@ if [ "$ISALIAS" = no ] && is_bonding_device ${DEVICE} ; then
 fi
 
 # If the device is part of a bridge, add the device to the bridge
-if [ -n "${BRIDGE}" ] && [ -x /usr/sbin/brctl ]; then
+if [ -n "${BRIDGE}" ]; then
     if [ ! -d /sys/class/net/${BRIDGE}/bridge ]; then
-        /usr/sbin/brctl addbr -- ${BRIDGE} 2>/dev/null
+        ip link add ${BRIDGE} type bridge 2>/dev/null
     fi
     /sbin/ip addr flush dev ${DEVICE} 2>/dev/null
     /sbin/ip link set dev ${DEVICE} up
     ethtool_set
     [ -n "${LINKDELAY}" ] && /bin/sleep ${LINKDELAY}
-    /usr/sbin/brctl addif -- ${BRIDGE} ${DEVICE}
+    ip link set ${DEVICE} master ${BRIDGE}
     # add the bits to setup driver parameters here
     for arg in $BRIDGING_OPTS ; do
         key=${arg%%=*};


network-scripts/ifdown-eth
```diff
@@ -135,14 +135,13 @@ if [ -d "/sys/class/net/${REALDEVICE}" ]; then
 fi
 [ "$retcode" = "0" ] && retcode=$?
 
-if [ -n "${BRIDGE}" ] && [ -x /usr/sbin/brctl ]; then
-    /sbin/ip link set dev ${DEVICE} down
-    /usr/sbin/brctl delif -- ${BRIDGE} ${DEVICE}
+if [ -n "${BRIDGE}" ]; then
```
erig0: Do we need to retain the "-x" check?
e.g.

    if [ -n "${BRIDGE}" ] && [ -x /usr/sbin/ip ]; then

network-scripts/ifdown-eth
```diff
@@ -135,14 +135,13 @@ if [ -d "/sys/class/net/${REALDEVICE}" ]; then
...
+    /sbin/ip link set dev ${DEVICE} nomaster down
     # Upon removing a device from a bridge,
     # it's necessary to make radvd reload its config
     [ -r /run/radvd/radvd.pid ] && kill -HUP $(cat /run/radvd/radvd.pid)
     if [ -d /sys/class/net/${BRIDGE}/brif ] && [ $(ls -1 /sys/class/net/${BRIDGE}/brif | wc -l) -eq 0 ]; then
```
erig0: It looks like this is making sure the bridge is not used by anything before attempting to delete it. Is there an ip-link equivalent that we can use? Maybe this works..

    if [ -z "$(ip link show master ${BRIDGE})" ]; then

network-scripts/ifup-eth
```diff
@@ -45,22 +45,20 @@ if [ -n "${HWADDR}" ]; then
...
-        net_log $"Bridge support not available: brctl not found"
-        exit 1
-    fi
+    OPTS=""
+    [ -n "${DELAY}" ] && OPTS+="forward_delay ${DELAY} "
+    is_true "${STP}" && OPTS+="stp 1 "
+    [ -n "${PRIO}" ] && OPTS+="priority ${PRIO} "
+    [ -n "${AGEING}" ] && OPTS+="ageing_time ${AGEING} "
 
     if [ ! -d /sys/class/net/${DEVICE}/bridge ]; then
```
erig0: ip-link can tell us if it exists. Does it make sense to check /sys?

network-scripts/ifup-eth
```diff
@@ -45,22 +45,20 @@ if [ -n "${HWADDR}" ]; then
...
-# If the device is a bridge, create it with brctl, if available.
+# If the device is a bridge, create it
 if [ "${TYPE}" = "Bridge" ]; then
-    if [ ! -x /usr/sbin/brctl ]; then
-        net_log $"Bridge support not available: brctl not found"
-        exit 1
-    fi
+    OPTS=""
+    [ -n "${DELAY}" ] && OPTS+="forward_delay ${DELAY} "
+    is_true "${STP}" && OPTS+="stp 1 "
```
erig0: Below we set these options even if the device already exists. So we need the "not true" case as well for STP so we disable STP on a device.

deekej: I think this is not correct. According to man page, only `stp_state` option exists for `ip link`.

network-scripts/ifdown-eth
```diff
@@ -135,14 +135,13 @@ if [ -d "/sys/class/net/${REALDEVICE}" ]; then
...
-        /usr/sbin/brctl delbr -- ${BRIDGE}
+        /sbin/ip link del ${BRIDGE}
```
deekej: Shouldn't this be `/sbin/ip link delete ${BRIDGE}` instead?

The `delbr` option of `brctl` is supposed to delete the `${BRIDGE}`. While the `del` option for `ip link` is supposed to remove MACADDR from allowed list. Are those really equivalent?

NOTE: The `delete` option for `ip link` is supposed to delete the virtual link, all according to man pages...

network-scripts/ifup-eth
```diff
@@ -45,22 +45,20 @@ if [ -n "${HWADDR}" ]; then
...
 fi
 
-# If the device is a bridge, create it with brctl, if available.
+# If the device is a bridge, create it
 if [ "${TYPE}" = "Bridge" ]; then
-    if [ ! -x /usr/sbin/brctl ]; then
-        net_log $"Bridge support not available: brctl not found"
-        exit 1
-    fi
+    OPTS=""
```
deekej: Please, just out of convention, use the `bridge_opts` variable name. (I'd like the keep the names in CAPITALS for variables set by user, and the lower caps for local variables. :)

network-scripts/ifup-eth
```diff
@@ -45,22 +45,20 @@ if [ -n "${HWADDR}" ]; then
...
+    is_true "${STP}" && OPTS+="stp 1 "
+    [ -n "${PRIO}" ] && OPTS+="priority ${PRIO} "
+    [ -n "${AGEING}" ] && OPTS+="ageing_time ${AGEING} "
 
     if [ ! -d /sys/class/net/${DEVICE}/bridge ]; then
-        /usr/sbin/brctl addbr -- ${DEVICE} || exit 1
+        /sbin/ip link add ${DEVICE} type bridge $OPTS
+    elif [ -n "${OPTS}" ]; then
+        /sbin/ip link set ${DEVICE} type bridge $OPTS
     fi
```
deekej: And just as a precaution, add the `unset bridge_opts` call here as well. Thanks! :)

n0-1:
Hi,

On Mon, Jun 11, 2018 at 08:24:30AM -0700, Eric Garver wrote:
> > @@ -135,14 +135,13 @@ if [ -d "/sys/class/net/${REALDEVICE}" ]; then
>  fi
>  [ "$retcode" = "0" ] && retcode=$?
>  
> -if [ -n "${BRIDGE}" ] && [ -x /usr/sbin/brctl ]; then
> -    /sbin/ip link set dev ${DEVICE} down
> -    /usr/sbin/brctl delif -- ${BRIDGE} ${DEVICE}
> +if [ -n "${BRIDGE}" ]; then
> 
> Do we need to retain the "-x" check?
> e.g.
> 
>     if [ -n "${BRIDGE}" ] && [ -x /usr/sbin/ip ]; then

I didn't have a close look, but it seems the code assumes existence of
'ip' tool. Without it, I guess it's not much use anyway.

But I noticed something different (maybe you actually had that in mind):
The existing code seems to arbitrarily choose from 'ip' and '/sbin/ip'
to refer to the binary. I will change my code to just use 'ip'
everywhere, at least that's what ifdown-eth is doing.

> > @@ -135,14 +135,13 @@ if [ -d "/sys/class/net/${REALDEVICE}" ]; then
>  fi
>  [ "$retcode" = "0" ] && retcode=$?
>  
> -if [ -n "${BRIDGE}" ] && [ -x /usr/sbin/brctl ]; then
> -    /sbin/ip link set dev ${DEVICE} down
> -    /usr/sbin/brctl delif -- ${BRIDGE} ${DEVICE}
> +if [ -n "${BRIDGE}" ]; then
> +    /sbin/ip link set dev ${DEVICE} nomaster down
>      # Upon removing a device from a bridge,
>      # it's necessary to make radvd reload its config
>      [ -r /run/radvd/radvd.pid ] && kill -HUP $(cat /run/radvd/radvd.pid)
>      if [ -d /sys/class/net/${BRIDGE}/brif ] && [ $(ls -1 /sys/class/net/${BRIDGE}/brif | wc -l) -eq 0 ]; then
> 
> It looks like this is making sure the bridge is not used by anything before attempting to delete it. Is there an ip-link equivalent that we can use? Maybe this works..
> 
>     if [ -z "$(ip link show master ${BRIDGE})" ]; then

I didn't want to mess with the old code too much. I guess your version
would work, but reading sysfs probably involves less overhead. Also we
don't have to depend upon ip-link output then. :)

> >  if [ "${TYPE}" = "Bridge" ]; then
> -    if [ ! -x /usr/sbin/brctl ]; then
> -        net_log $"Bridge support not available: brctl not found"
> -        exit 1
> -    fi
> +    OPTS=""
> +    [ -n "${DELAY}" ] && OPTS+="forward_delay ${DELAY} "
> +    is_true "${STP}" && OPTS+="stp 1 "
> +    [ -n "${PRIO}" ] && OPTS+="priority ${PRIO} "
> +    [ -n "${AGEING}" ] && OPTS+="ageing_time ${AGEING} "
>  
>      if [ ! -d /sys/class/net/${DEVICE}/bridge ]; then
> 
> ip-link can tell us if it exists. Does it make sense to check /sys?

Same as above.

> >  if [ "${TYPE}" = "Bridge" ]; then
> -    if [ ! -x /usr/sbin/brctl ]; then
> -        net_log $"Bridge support not available: brctl not found"
> -        exit 1
> -    fi
> +    OPTS=""
> +    [ -n "${DELAY}" ] && OPTS+="forward_delay ${DELAY} "
> +    is_true "${STP}" && OPTS+="stp 1 "
> 
> Below we set these options even if the device already exists. So we need the "not true" case as well for STP so we disable STP on a device.

Oh, good catch! I'll fix that in v2.

Thanks, Phil

n0-1:
Hi,

On Mon, Jun 11, 2018 at 08:29:32AM -0700, deekej wrote:
> >      # Upon removing a device from a bridge,
>      # it's necessary to make radvd reload its config
>      [ -r /run/radvd/radvd.pid ] && kill -HUP $(cat /run/radvd/radvd.pid)
>      if [ -d /sys/class/net/${BRIDGE}/brif ] && [ $(ls -1 /sys/class/net/${BRIDGE}/brif | wc -l) -eq 0 ]; then
> -        /usr/sbin/brctl delbr -- ${BRIDGE}
> +        /sbin/ip link del ${BRIDGE}
> 
> Shouldn't this be `/sbin/ip link delete ${BRIDGE}` instead?
> 
> The `delbr` option of `brctl` is supposed to delete the `${BRIDGE}`. While the `del` option for `ip link` is supposed to remove MACADDR from allowed list.
> 
> The `delete` option for `ip link` is supposed to delete the virtual link.
> 
> ... All according to man pages.

That's not correct. What you are referring to is 'macaddr del' option of
'ip link set' command. For plain 'ip link', 'del' is just short for
'delete'. But I can change that to avoid confusion.

Cheers, Phil

network-scripts/ifup-eth
n0-1: Crap, you found the typo I knew was hiding somewhere. :)

Thanks!

network-scripts/ifup-eth
n0-1: OK, will do. Same for 'unset'.

network-scripts/ifup-eth
deekej: If you want to make this change, then I would suggest fixing it everywhere in network-scripts, and doing a separate commit for it.

network-scripts/ifup-eth
deekej: Good catch! :)

deekej:
> That's not correct. What you are referring to is 'macaddr del' option of 'ip link set' command. For plain 'ip link', 'del' is just short for 'delete'. But I can change that to avoid confusion.

Ah, okay. :) Thanks for the explanation! ;)

n0-1:
Sending a v2.

network-scripts/ifup-eth
```diff
@@ -173,15 +176,15 @@ if [ "$ISALIAS" = no ] && is_bonding_device ${DEVICE} ; then
...
 
 # If the device is part of a bridge, add the device to the bridge
-if [ -n "${BRIDGE}" ] && [ -x /usr/sbin/brctl ]; then
+if [ -n "${BRIDGE}" ]; then
     if [ ! -d /sys/class/net/${BRIDGE}/bridge ]; then
-        /usr/sbin/brctl addbr -- ${BRIDGE} 2>/dev/null
+        ip link add ${BRIDGE} type bridge 2>/dev/null
     fi
     /sbin/ip addr flush dev ${DEVICE} 2>/dev/null
     /sbin/ip link set dev ${DEVICE} up
```
keszybz: You should probably drop `/sbin/` here to match the other calls to `ip`.

Issue #210, opened 2018-06-07
Remove (deprecated) brctl calls with ip-link


deekej:
More info: https://bugzilla.redhat.com/show_bug.cgi?id=1588180

deekej:
Resolved with #213.

PR #213, opened 2018-06-11
network-scripts: Replace brctl with ip-link (v2)

n0-1:
Since ip-link has full support for Linux bridges (and slave ports), use
that instead of the deprecated brctl from bridge-utils.

Changes since v1:
- Don't call 'ip' by full path, let the shell find it.
- Use lower-case 'bridge_opts' variable name instead of upper-case 'OPTS' and unset it after use.
- Fix typo: 'stp' -> 'stp_state'.
- Fix for 'STP=no' to disable spanning-tree if the bridge already exists.


network-scripts/ifup-eth
```diff
@@ -45,21 +45,24 @@ if [ -n "${HWADDR}" ]; then
...
+	    bridge_opts+="stp_state 1 "
+    elif is_false "${STP}"; then
+	    bridge_opts+="stp_state 0 "
     fi
+    [ -n "${PRIO}" ] && bridge_opts+="priority ${PRIO} "
+    [ -n "${AGEING}" ] && bridge_opts+="ageing_time ${AGEING} "
 
     if [ ! -d /sys/class/net/${DEVICE}/bridge ]; then
-        /usr/sbin/brctl addbr -- ${DEVICE} || exit 1
+        ip link add ${DEVICE} type bridge $bridge_opts
```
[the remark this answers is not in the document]
deekej: Hmm, short answer is - yes, lets keep the `|| exit 1`.

In case the ip link can fail here as well, then we should exit if this happens. This is just for the reason to not break already existin setups. If somebody will find it problematic in the future, then they'll need to open BZ/issue for it, so we can better understand the reason and to better track it.

network-scripts/ifup-eth
```diff
@@ -45,21 +45,24 @@ if [ -n "${HWADDR}" ]; then
...
+    elif [ -n "${OPTS}" ]; then
+        ip link set ${DEVICE} type bridge $bridge_opts
```
deekej: `|| exit 1` -- same here as well I would say.

n0-1: DONE! :)

deekej:
Resolves: #210

PR #220, opened 2018-06-19
Move the /etc/rwtab.d & /etc/statetab.d folders to 'filesystem'

deekej:
This will allow services to install drop-in files into these folders for the 'readonly-root' service, but that service won't have to be installed by default as a result...

NOTE: We have to wait until the https://pagure.io/filesystem/pull-request/1 is merged, before we can merge the change in initscripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,8 +85,6 @@ install-man: install-usr
 # Initscripts still ship some empty directories necessary for system to function
 # correctly...
 install-post: install-etc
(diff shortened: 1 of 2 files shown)


deekej:
I'd rather not separate the drop-in folders, if we are moving them somewhere else. This should allow administrator to install custom configuration to /etc on fresh machine before downloading & installing of necessary packages (`readonly-root` in this case).

PR #228, opened 2018-07-30
ifup-eth: use 'bc' instead of 'expr' when computing $forward_delay

deekej:
Because the return value of `convert2sec()` function can sometimes be decimal, the follow up `expr` call can fail, since `expr` does not support floating point calculations. This can sometimes lead to error:

```shell
expr: non-integer argument
/etc/sysconfig/network-scripts/ifup-eth: line 91: [: 0: unary operator expected
```

To solve this bug, we switch to `bc` utility, which supports floating point computations. Using `bc` should also be much faster than calling `gawk` to do this.

Resolves: #227

### DIFF
--- a/initscripts.spec
+++ b/initscripts.spec
@@ -94,6 +94,7 @@ Requires:         %{name}%{?_isa} = %{version}-%{release}
 
 %shared_requirements
 
+Requires:         bc
 Requires:         dbus
 Requires:         gawk
 Requires:         grep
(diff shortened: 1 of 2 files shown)


vfreex:
This actually doesn't work either, because the single square bracket `[ ]` only supports integers for comparison. I would suggest rounding up the `forward_delay` value instead of dealing with floating numbers.

BTW, bug created https://bugzilla.redhat.com/show_bug.cgi?id=1609687

deekej:
@vfreex Thanks for letting me know, I'll check what can be done about it. :) And thanks for the bug report! :)
@lnykryn You're right. I lived thinking `bc` was always part of `coreutils`. But I think adding the dependency on `bc` should be okay, as well doing computations with it. IIRC, it should be much faster compared to gawk, etc.

PR #245, opened 2019-03-05
ifup-eth: add a new PERSISTENT_DHCLIENT_IPV6 option for IPv6 dhclient…

ZhiqiangLiu26:
In 76226a34 ("ifup-eth: apply PERSISTENT_DHCLIENT to IPv6 dhclient daemon"),
one PERSISTENT_DHCLIENT option works for both IPv4 and IPv6, so the IPv4
and IPv6 settings are consistent. However, the user settings for IPv4 and IPv6
are limited.

For example, users try to adopt both IPv4 and IPv6 protocol. DHCPv6 servers
may be not stable as DHCP servers, thus the expectation for obtaining IPv6
address is lower than IPv4 address. For users, IPv4 address must be
obtained for the basic communication. So the PERSISTENT_DHCLIENT option is
set to be "yes | 1" for IPv4 address. However, the network service may be
stuck until it fails due to missing DHCPv6 servers, and the remaining devices also
cannot obtain both IPv4 and IPv6 addresses. The problem is very serious for users.

Here, I add a new PERSISTENT_DHCLIENT_IPV6 option only for IPv6 dhclient daemon,
so users can set IPv4 and IPv6 separately.

Fixes: 76226a3 ("ifup-eth: apply PERSISTENT_DHCLIENT to IPv6 dhclient daemon")
Signed-off-by: Zhiqiang Liu <liuzhiqiang26@huawei.com>

### DIFF
--- a/doc/sysconfig.txt
+++ b/doc/sysconfig.txt
@@ -586,6 +586,15 @@ Files in /etc/sysconfig/network-scripts/
       By default network-scripts will set `accept_ra` only if $IPV6_AUTOCONF is
       set to `yes`. If you don't want SLAAC addresses but do want to accept RA,
       then set this to `yes`. Defaults to `no`.
+    PERSISTENT_DHCLIENT_IPV6=yes|no|1|0
+      Without this option, or if it is 'no'/'0', and DHCPV6C=yes,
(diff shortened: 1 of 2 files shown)


jamacku (approved):
LGTM

PR #246, opened 2019-03-14
Add option to wait until target is reachable for network start command

jamacku:


etc/rc.d/init.d/network
```diff
@@ -164,15 +164,16 @@ start)
     touch /var/lock/subsys/network
 
     # Fixed delay
-    [ -n "${NETWORKDELAY}" ] && [ -z "${MINIMALDELAY}" ] || is_false "${MINIMALDELAY}" && /bin/sleep "${NETWORKDELAY}"
+    [ -n "${NETWORKDELAY}" ] && [ -z "${NETWORKTESTSERVER}" ] && /bin/sleep "${NETWORKDELAY}"
     
     # Adaptive delay
     # It tests if network connection is ready via ping
-    if is_true "${MINIMALDELAY}" ; then
+    if [ -n "${NETWORKTESTSERVER}" ] ; then
```
lnykryn: I would probably name this WAIT_UNTIL_REACHABLE

Also an you please document the option in doc/sysconfig.txt (under NETWORKDELAY)

etc/rc.d/init.d/network
```diff
@@ -164,15 +164,16 @@ start)
...
       [ -z "${NETWORKDELAY}" ] && NETWORKDELAY=30
-      SYSTEMTIME=$(date "+%s")
-      while [ "$(date "+%s")" -le $(( SYSTEMTIME + NETWORKDELAY )) ] ; do
-        ping -q -c 1 -W 1 8.8.8.8 >/dev/null && break
+      TIMEOUT=0
+      while [ $TIMEOUT -le $NETWORKDELAY ] ; do
+        ping -q -c 1 -W 1 "${NETWORKTESTSERVER}" >/dev/null && break
```
lnykryn: Maybe we should log if we are not able to ping the address, but we already reached the timeout. Something like  
net_log $"Target IP is not reachable, but we already reached the timeout. Continuing anyway". But I am not sure if it is needed, I will leave that up to you.

etc/rc.d/init.d/network
jamacku: I renamed variable as you suggested and I added documentation.

etc/rc.d/init.d/network
jamacku: I think that it is really good idea. I added log as you suggested. Thank you.

etc/rc.d/init.d/network
```diff
@@ -163,7 +163,20 @@ start)
...
+    # Adaptive delay
+    # It tests if network connection is ready via ping
+    if [ -n "${WAIT_UNTIL_REACHABLE}" ] ; then
+      [ -z "${NETWORKDELAY}" ] && NETWORKDELAY=30
+      TIMEOUT=0
+      while [ $TIMEOUT -le $NETWORKDELAY ] ; do
+        ping -q -c 1 -W 1 "${WAIT_UNTIL_REACHABLE}" >/dev/null && break
+        TIMEOUT=$(( TIMEOUT + 1 ))
+      done
+      net_log $"Target IP is not reachable, but timeout was already reached. Continuing anyway."
```
lnykryn: This will cause that the message will be print every time. You probably want to add 
[ $TIMEOUT -eq $NETWORKDELAY ] && net_log ...

jamacku: Fixed

etc/rc.d/init.d/network
```diff
@@ -163,7 +163,22 @@ start)
...
+    if [ -n "${WAIT_UNTIL_REACHABLE}" ] ; then
+      [ -z "${NETWORKDELAY}" ] && NETWORKDELAY=30
+      TIMEOUT=0
+      IPV=""
+      ipcalc -c -6 "${WAIT_UNTIL_REACHABLE}" &>/dev/null && IPV="-6"
+      while [ $TIMEOUT -le $NETWORKDELAY ] ; do
+        ping $IPV -q -c 1 -W 1 "${WAIT_UNTIL_REACHABLE}" >/dev/null && break
+        TIMEOUT=$(( TIMEOUT + 1 ))
+      done
+      [ $TIMEOUT -ge $NETWORKDELAY ] && net_log $"Target IP is not reachable, but timeout was already reached. Continuing anyway."
```
lnykryn: Since the loop is while [ $TIMEOUT -le $NETWORKDELAY ]  the same value of TIMEOUT and NETWORKDELAY is still fine. So I think here it should be  -gt instead of -ge. Right?

jamacku: You are right. Thank you.

PR #247, opened 2019-03-27
Remove remaining bashisms in /etc/rc.d/init.d/functions

jccleaver:
Replaces insecure bash $"..." constructs with gettext/eval_gettext calls
Removes hard-coded /bin/bash for sysconfig-overrideable /bin/sh in daemon()
Rewrites prior ksh loading patch to also be executable with non-bash: 
https://github.com/fedora-sysv/initscripts/commit/8cb57357a84900707d25c8f74184891122713df0

### DIFF
--- a/etc/rc.d/init.d/functions
+++ b/etc/rc.d/init.d/functions
@@ -291,27 +291,14 @@ daemon() {
     # if they set NICELEVEL in /etc/sysconfig/foo, honor it
     [ -n "${NICELEVEL:-}" ] && nice="nice -n $NICELEVEL"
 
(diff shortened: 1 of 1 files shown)


jamacku:
Hello,
This is in theory a good idea, but I'm a bit sceptical about this. Initscripts are for a long time in maintenance mode only and this change could potentially break current scripts, cause some regressions or even introduce new bugs.

Also for example Fedora has as default shell bash. I don't want to defend bashisms in Initscripts, but there is more risk involved then potential reward. Thank you for understanding.

I will close this PR. Thank you for your time and patience. 
Jan

PR #263, opened 2019-07-04
all: source /functions from the correct location

lkundrak:
The functions file is actually installed into
/etc/rc.d/init.d/functions, not /etc/init.d/functions.

The symlink from /etc/init.d to /etc/rc.d/init.d is provided by the
chkconfig package, but it's not always there.

### DIFF
--- a/etc/rc.d/init.d/network
+++ b/etc/rc.d/init.d/network
@@ -14,7 +14,7 @@
 ### END INIT INFO
 
 # Source function library.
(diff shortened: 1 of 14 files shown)


lnykryn:
We fixed this the other way around and now the init script package creates the symlink
https://github.com/fedora-sysv/initscripts/commit/fe85d7bfa8909059716d17802b89454dfb8f9f19

PR #268, opened 2019-07-22
ifup/ifdown: print DEPRECATION_WARNING_ISSUED waring info after source_config

ZhiqiangLiu26:
In ifup/ifdown scripts, move deprecation waring info after
source_config, so users can config DEPRECATION_WARNING_ISSUED in
ifcfg-** file to decide whether show deprecation waring info when
calling ifup/ifdown.

Signed-off-by: Zhiqiang Liu <liuzhiqiang26@huawei.com>

### DIFF
--- a/network-scripts/ifdown
+++ b/network-scripts/ifdown
@@ -16,12 +16,6 @@ CONFIG=$1
     exit 1
 }
 
-if ! is_true ${DEPRECATION_WARNING_ISSUED}; then
-    net_log $"You are using 'ifdown' script provided by 'network-scripts', which are now deprecated." warning ifdown >&2
-    net_log $"'network-scripts' will be removed from distribution in near future." warning ifdown >&2
-    net_log $"It is advised to switch to 'NetworkManager' instead - it provides 'ifup/ifdown' scripts as well." warning ifdown >&2
-fi
-
 need_config "${CONFIG}"
 
 [ -f "$CONFIG" ] || {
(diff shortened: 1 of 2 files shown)


deekej:
I'm sorry, but I don't see this as a good idea. The `DEPRECATION_WARNING_ISSUED` is meant for internal usage of initscripts, IIRC, so that deprecation warning are displayed once (not multiple times).

I really don't think that we want users of Fedora to easily shut down deprecation warnings. That usually means they will not switch to new way of doing things. Once the network-scripts will get removed, then it will cause unexpected troubles, because they will most likely forget that they have put something in their ifcfg-* files (they are not updated by rpm).

I'm against this PR, but the ultimate decision is on @lnykryn, since he's the current maintainer of initscripts.

ZhiqiangLiu26:
> I'm sorry, but I don't see this as a good idea. The `DEPRECATION_WARNING_ISSUED` is meant for internal usage of initscripts, IIRC, so that deprecation warning are displayed once (not multiple times).
> 
> I really don't think that we want users of Fedora to easily shut down deprecation warnings. That usually means they will not switch to new way of doing things. Once the network-scripts will get removed, then it will cause unexpected troubles, because they will most likely forget that they have put something in their ifcfg-* files (they are not updated by rpm).
> 
> I'm against this PR, but the ultimate decision is on @lnykryn, since he's the current maintainer of initscripts.

Thanks for your reply. You are mostly right.
I have the PR just because that some users are used to using network.service and ifup/ifdown to manage the network devices. If the deprecation warnings are shown every time, it may bring users a bad experience. So DEPRECATION_WARNING_ISSUED could give users a chance to config whether show deprecation waring info when calling ifup/ifdown.
Thanks again.

lnykryn:
Looks good to me

PR #286, opened 2019-09-23
Stop disabling IPV6 by default

zezinho42:
With this default setting, when changing from one Wifi network to another, IPV6 was always disabled, even if it was enabled in the source network-script file.


lnykryn:
I don't think this change does what you want, this is only related to the ipv4 aliases.

jamacku:
Feel free to reopen this PR in case of any questions.

Issue #300, opened 2020-01-20
grep -L return value breaks network start


dirteat:
Spotted by M. W. on Mageia:

* Noteworthy changes in release 3.2 (2018-12-20) [stable]
** Changes in behavior
  The --files-without-match (-L) option now causes grep to succeed
  when a file is listed, instead of when a line is selected.  This
  resembles what git-grep does.

This breaks the init.d/network scripts there:

`if LANG=C grep -EL "^ONBOOT=['\"]?[Nn][Oo]['\"]?" ifcfg-$i > /dev/null ;`
`then`
`     # this loads the module, to preserve ordering`
`          is_available $i`
`          continue`
`fi`

and just after for the `# Bring up xDSL and VPN interfaces`.

The fix is to either use a negation, or the negated option `-l`

Cheers,
Chris.

lnykryn:
I would suggest replacing this with something like
if ( . ./ifcfg-$i ; is_false $ONBOOT ;  ) ; then ....
Using standard is_true/is_false function but sourcing the ifcfg file in the subshell, so we don't mess up the current environment.

jamacku:
Hi,
Pull request #301 should fix your issue. In case It won't, please let me know.

jamacku:
Fixed in initscripts 10.03

PR #301, opened 2020-01-24
Replace grep -EL with subshell since -L changed behaviour

jamacku:
Sugested by @lnykryn
Related to issue #300


jamacku:
CI issue

jamacku:
CI is failing since path in command is dynamic and shellcheck can't follow it.

PR #302, opened 2020-01-24
Change path to arping binary

jamacku:
Resolves bz #1773196


lnykryn:
@jsynacek could you please review this? Iputils package is your field of expertise ;-)

jamacku:
> /sbin/arping has been moved to /usr/sbin/arping. And because /sbin is a symlink to /usr/sbin, there is no reason why scripts that call /sbin/arping should stop working.

Thank you @jsynacek .
I'm closing this PR since this isn't bug in initscripts.

Issue #42, opened 2016-11-16
Use is_true() & is_false() everywhere


deekej:
Currently, there are still some places where there are hardcoded tests for `yes` and `no` values, instead of using `is_true` and `is_false` functions.

Example: https://github.com/fedora-sysv/initscripts/blob/master/sysconfig/network-scripts/network-functions#L251

For more sane approach, we should no longer use this parsing.

jamacku:
#242

PR #312, opened 2020-03-05
Use function is_true for testing true conditions

jamacku:
Related to #42


jamacku:
It looks like CI works :)

PR #315, opened 2020-03-11
 Wait for scope link addresses as well as for scope global addresses

jamacku:
Fix issue when interface has assigned only local address in tentative state and DHCPv6 isn't able to assign address to that interface. With this patch network-scripts will wait until tentative state is gone (wait for DAD) for all scopes including local one.

Resolves: #1809601


jamacku:
> I see what the code is doing but commit message doesn't explain why. Please be elaborate and explain the motivation "why" we need this change now.

Thank you, Michal, good point. I updated commit message.

PR #323, opened 2020-05-17
Correct spelling, IP, …

comradekingu:


comradekingu:
@jamacku It isn't typographically correct to use tripledot. What's worse it will spell out "dot-dot-dot" for people that use screen-readers. How it looks is a concern with the font it is rendered in I should think.

jamacku:
@comradekingu Thank you for clarification. I'll merge your PR's once I'll be doing new release.

PR #328, opened 2020-07-02
rename_device: also support dracut-style kernel cmdline configuration

lnykryn:
There are 2 cases where this might be useful.
* On machines with initrd build without dracut-network, you can now
  still use ifname option
* On older kernels where double renaming was allowed, normal system
  replaced the name set by initrd.

### DIFF
--- a/src/rename_device.c
+++ b/src/rename_device.c
@@ -278,6 +278,46 @@ char *get_config_by_hwaddr(char *hwaddr, char *current) {
 	return first;
 }
 
+/* Let's check kernel cmdline and also process ifname= entries
+ * as they are documented in dracut.cmdline(7)
+ * Example: ifname=test:aa:bb:cc:dd:ee:ff
+ */
+char *get_cmdline_by_hwaddr(char *hwaddr) {
+	gchar *contents;
+	gchar **entries;
+	int i;
+	char *name = NULL;
+	g_file_get_contents("/proc/cmdline", &contents, NULL, NULL);
+	entries = g_strsplit(contents," ", 0);
+	for (i = 0; entries[i]; i++) {
+		char *c = (char *) entries[i];
+		char *n;
+
+		if (!g_str_has_prefix((gchar *)c,"ifname="))
+			continue;
+
+		c = strstr(c, "=");
+		if (c == NULL)
+			continue;
+		c++;
+		n=c;
+
+		c = strstr(c, ":");
+		if (c == NULL)
+			continue;
+		*c = '\0';
+		c++;
+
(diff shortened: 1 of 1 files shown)


lnykryn:
@jamacku @msekletar  What do you think? https://bugzilla.redhat.com/show_bug.cgi?id=1851503

src/rename_device.c
```diff
@@ -278,6 +278,41 @@ char *get_config_by_hwaddr(char *hwaddr, char *current) {
...
+	gchar **entries;
+	int i;
+	char *name = NULL;
+	g_file_get_contents("/proc/cmdline", &contents, NULL, NULL);
+	entries = g_strsplit(contents," ", 0);
+	for (i = 0; entries[i]; i++) {
+		gchar *c = entries[i];
+		gchar **pair;
+
+		if (!g_str_has_prefix(c,"ifname="))
```
jamacku, line 296: It looks to me like this will continue at all times, or am I missing something? Thanks.

src/rename_device.c
```diff
@@ -278,6 +278,46 @@ char *get_config_by_hwaddr(char *hwaddr, char *current) {
...
+	g_file_get_contents("/proc/cmdline", &contents, NULL, NULL);
+	entries = g_strsplit(contents," ", 0);
+	for (i = 0; entries[i]; i++) {
+		char *c = (char *) entries[i];
+		char *n;
+
+		if (!g_str_has_prefix((gchar *)c,"ifname="))
+			continue;
+
+		c = strstr(c, "=");
```
dtardon, line 299: We already know that `c` starts with `"ifname="`, so there's no need to search anyting. A simple increment is sufficient.

src/rename_device.c
```diff
@@ -278,6 +278,46 @@ char *get_config_by_hwaddr(char *hwaddr, char *current) {
...
+		if (c == NULL)
+			continue;
+		c++;
+		n=c;
+
+		c = strstr(c, ":");
```
dtardon, line 305: Why not `strchr`?

src/rename_device.c
```diff
@@ -278,6 +278,46 @@ char *get_config_by_hwaddr(char *hwaddr, char *current) {
...
+		if (c == NULL)
+			continue;
+		*c = '\0';
+		c++;
```
dtardon, line 309: This is not needed.

jamacku:
@dtardon Thank you for review.

Rename-devices will be reworked in rust, so I'll merge this one, so we won't forgot about this feature.

jamacku (approved):
LGTM, with some minor issues which will be mitigated in rework.

PR #331, opened 2020-07-13
Execute udevadm settle

zhangnaru0605:
Execute udevadm settle to wait for all udev events to be handled when starting network


lnykryn:
In most cases just calling udevadm settle is not a good idea. It delays the boot for a significant time and does not guarantee much. If you have a setup where network card can appear anytime, it is better to use NetworkManager, that is designed for that.

lnykryn:
Btw, you can always call 
# systemctl add-wants network-pre.target systemd-udev-settle.service

zhangnaru0605:
Thanks for your suggestion!

PR #334, opened 2020-07-24
Add optional 'dev' keyword

jamacku:
Fix the problem when the device name could be interpreted as an iproute2 keyword. For example, for a bridge slave named "a" the iproute2 would treat the name as a prefix of keyword "address" and the network-scripts would fail to set the bridge master.

Related: rhbz #1859785

### DIFF
--- a/network-scripts/ifup
+++ b/network-scripts/ifup
@@ -140,7 +140,7 @@ if is_true "${VLAN}" && is_false "$ISALIAS" && [ -n "$DEVICE" ]; then
                 exit 1
             }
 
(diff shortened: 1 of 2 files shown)


jamacku:
@msekletar, Could you please review this. Thank you.

msekletar:
LGTM, however the commit message is very vague (specifically part `Specific device names could be misinterpreted as something else`) so I had to look up some details in the corresponding BZ.

msekletar:
Suggestion for better commit message (based on BZ problem description)


Add optional 'dev' keyword

Fix the problem when the device name could be interpreted as an iproute2 keyword. For example, for a bridge slave named "a" the iproute2 would treat the name as a prefix of keyword "address" and the network-scripts would fail to set the bridge master.

jamacku:
Thank you @msekletar.

Issue #337, opened 2020-08-11
initscripts.spec: network-scripts package should Require: ethtool


ovitters:
`network-scripts/ifup-eth` calls the function `ethtool_set` various times, see e.g.:
https://github.com/fedora-sysv/initscripts/blob/fc6164b75090f03eed7504195d1fa9f0915ea862/network-scripts/ifup-eth#L144-L149

this `ethtool_set` function is defined in `network-scripts/network-functions`, see 

https://github.com/fedora-sysv/initscripts/blob/fc6164b75090f03eed7504195d1fa9f0915ea862/network-scripts/network-functions#L198-L217

However, initscripts.spec does not have a require for ethtool. I think it should be added in the network-scripts package.

Current Requires are:

https://github.com/fedora-sysv/initscripts/blob/fc6164b75090f03eed7504195d1fa9f0915ea862/initscripts.spec#L97-L108

I do see in the history that various changes were made to reduce the ethtool dependency/usage.

jamacku:
Thank you for reporting this issue.

jamacku:
Fixed

PR #338, opened 2020-08-12
Add ethtool into required packages for network-scripts - Olav Vitters

jamacku:
Fixes issue #337


lnykryn:
To be honest I am not a fan of adding this as a hard dependency since not a lot of people are using this option.
I think we should add a warning if ETHTOOL_OPTS is set and the binary is not on the system.

jamacku:
Thanks @lnykryn, good point. Please could you review changes?

PR #340, opened 2020-08-13
rc.d/functions: replace grep's --quiet with -q

chaitu236:
Some versions of grep (like busybox's) do not support --quiet.
So use -q instead like in the rest of the file.

Signed-off-by: Chaitanya Vadrevu <chaitanya.vadrevu@ni.com>


lnykryn:
What is the real use-case here?

chaitu236:
We at NI use this file in our Linux RT distribution. The version of grub we use does not support "--quiet" but does support "-q". Hence the change.
FYI all other grub invocations in this file already use "-q" so this change is only bringing parity.

PR #345, opened 2020-09-24
Allow updating rfkill switch status while in readonly root mode

jamacku:
Adding rfkill switch status into rwtab.

Resolves: #1878874

(cherry picked from commit 537a39f10173d34cd9a2cf75e42225bb339c318e)


jamacku:
Blocked by #343

PR #347, opened 2020-10-20
init.d/functions: Make usage msgs more clear

jamacku:


jamacku:
CI Output:
```
E: Unable to locate package csbuild

The command "sudo apt-get install -y --force-yes csbuild" failed and exited with 100 during .
```

etc/rc.d/init.d/functions
```diff
@@ -448,7 +448,7 @@ status() {
     if [ "$1" = "-b" ]; then
         if [ -z $pid_file ]; then
             echo $"-b option can be used only with -p"
-            echo $"Usage: status -p pidfile -b binary program"
+            echo $"Usage: status [-p {pidfile} [-b {binary}]] [-l lockfile] {program}"
```
lnykryn, line 451: Looking at the code above, this does not seem correct. Currently we parse it in the -p -l- b order

lnykryn (changes requested):

etc/rc.d/init.d/functions
jamacku: Yes, you are right, thank you.

jamacku: Actually, I wanted to somehow express this relation:
``"-b option can be used only with -p"``

lnykryn (approved):

PR #353, opened 2020-11-11
service: catch unsupported action keywords

jamacku:
https://github.com/systemd/systemd/blob/master/src/systemctl/systemctl.c#L957-L1031

Resolves: #1894492


keszybz:
LGTM.

PR #355, opened 2020-12-01
network: add option to keep the link down

jamacku:
Some interfaces like Open vSwitch bridges using the userspace datapath,
needs the link to be kept down to avoid issues.
Add a LINKSTATUS=[down|up] ifcfg parameter to add this functionality.
If not specified, the link will default to up as before.

Patch provided by Matteo Croce mcroce@redhat.com

Resolves: #1555001


jamacku:
Original PR: #175

lnykryn (approved):

PR #359, opened 2021-01-04
remove rename_device_lock when process does not exist

wangxp006:
if rename_device is killed during fopen period in take_lock ,LOCKFILE will not be remove in sighandler.
Excute rename_device again,  LOCKFILE  is exist ,but the process with pid in LOCKFILE  is not running, take_lock will never return.


jamacku (approved):
LGTM

jamacku:
@wangxp006 Thank you for patch.

PR #369, opened 2021-02-12
 network scripts: Avoid infinite loop of arping

jamacku:
Introduced in the bonding driver (commit ae46f184bc1f) Driver now reports transmission failures.
Before that, it silently dropped the packet and replied with success error code.

The arping of iputils retries endlessly when a transmission error occurs.

This patch fix this behavior.


jamacku:
CI fails due to missing csbuild in testsuite

lnykryn (approved):

PR #372, opened 2021-03-10
ci: Migrate from Travis to GH Actions

jamacku:
There are two workflows, integration_test and shellcheck_test.
Those are similar to original travis jobs.


.github/workflows/check-shell.sh
```diff
@@ -1,21 +1,25 @@
 #!/bin/bash
 
-# Path is wrong because of travis
-. ./.ci/functions.sh
+PWD=".github/workflows"
```
mrc0mmand, line 3: This is a bit risky, since `$PWD` gets overwritten with every `cd`, e.g.:

```bash
#!/bin/bash -x

echo "$PWD"
PWD="test"
echo "$PWD"
cd /tmp
echo "$PWD"
```

```
$ ./test.sh
+ echo /home/mrc0mmand/tmp/bash
/home/mrc0mmand/tmp/bash
+ PWD=test
+ echo test
test
+ cd /tmp
+ echo /tmp
/tmp
```

It may work, in this case, since there are no `cd`s involved, but it may cause unexpected issues in the future.

Also, you can *detect* the script's working directory by using something like:

```bash
SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
```

so you don't have to figure out the path manually each time you shuffle the scripts around.

.github/workflows/check-shell.sh
```diff
@@ -1,21 +1,25 @@
...
 # get names of files from PR (excluding deleted files)
-git diff --name-only --diff-filter=db "${TRAVIS_COMMIT_RANGE}" > ../pr-changes.txt
+git diff --name-only --diff-filter=db "$git_base".."$git_head" > ../pr-changes.txt
 
 # Find modified shell scripts
 list_of_changes=()
 file_to_array "../pr-changes.txt" "list_of_changes" 0
 list_of_scripts=()
-file_to_array "./.ci/script-list.txt" "list_of_scripts" 1
+file_to_array "./${PWD}/script-list.txt" "list_of_scripts" 1
```
mrc0mmand, line 22: With the proposed changes above this can be then changed to:

```bash
file_to_array "$SCRIPT_DIR/script-list.txt" "list_of_scripts" 1
```

.github/workflows/check-shell.sh
```diff
@@ -27,7 +31,7 @@ done
 
 # Get list of exceptions
 list_of_exceptions=()
-file_to_array "./.ci/exception-list.txt" "list_of_exceptions" 1
+file_to_array "./${PWD}/exception-list.txt" "list_of_exceptions" 1
```
mrc0mmand, line 34: ditto

mrc0mmand (changes requested):

.github/workflows/check-shell.sh
jamacku: That is a good point. Thank you.

jamacku:
Test of CI: https://github.com/jamacku/initscripts/pull/34/checks

PR #373, opened 2021-04-12
spec: Move service script into subpackage

jamacku:
This changes moves service script into `initscripts-service` subpackage. The subpackage
could be install without main package.

This change allows `audit` not to require `initscripts` and require only `initscripts-service` instead.

Requested by: msekleta@redhat.com


msekletar (approved):
LGTM

Maybe just change Requires to Recommends dep from the main package.

jamacku:
Thank you for review, Michal.

PR #374, opened 2021-04-15
spec: Initscripts now requires chkconfig

jamacku:
As a easy way to get full legacy system V support.

### DIFF
--- a/initscripts.spec
+++ b/initscripts.spec
@@ -34,6 +34,7 @@ Requires:         procps-ng
 Requires:         setup
 Requires:         systemd
 Requires:         util-linux
+Requires:         chkconfig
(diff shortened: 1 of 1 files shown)


jamacku:
CI: `bash: ./.github/workflows/check-shell.sh: No such file or directory`

No shell script was changed.

jamacku:
@AdamWill, In my opinion initscripts doesn't have to be in default Fedora's network install images.

AdamWill:
The reason for its inclusion is: "## import-state.service for switchroot" (added in 2018). `/usr/lib/systemd/system/import-state.service` is still a part of initscripts, so it looks like we probably still need it.

PR #392, opened 2021-09-15
shellcheck: Fix shellcheck warnings

jamacku:
doc: [ShellCheck](https://github.com/koalaman/shellcheck/wiki)


jamacku:
@Mergifyio refresh

jamacku:
Let's don't risk this, when there is Differential ShellCheck in place...

Issue #396, opened 2021-10-21
killproc() does not execute properly on a busybox-bash environment


mengtanhzc:
**initscripts version the issue has been seen with**
1) initscripts 10.06
2) BusyBox v1.32.1

**Steps to reproduce the problem**
```
    #!/bin/bash
    
    . /etc/init.d/functions
    
    test()
    {
            sleep 30 &
            killproc sleep
    }
    
    test
```
run this testcase returns an error, and sleep is not killed.
![image](https://user-images.githubusercontent.com/41482086/138225687-eb9140ee-5949-4a3f-ad4c-5e7539b3b66f.png)

Instead of using bash package on my environment, I am using busybox-bash, but **busybox-bash does not support +=**, thus causing this problem. 
Can I submit a PR to fix this problem?

jamacku:
Feel free to submit a PR I'll be happy to review it.

mengtanhzc:
> Feel free to submit a PR I'll be happy to review it.

Thank you very much! I have just created a PR to solve it, please review~

jamacku:
Thank you for your contribution!

PR #397, opened 2021-11-01
Make the init.d/network and init.d/functions available on Busybox's shell

mengtanhzc:
#396 
In some embedded environments, instead of using GNU Bash for size reasons, the Ash (shell provided by busybox) is used. But init.d/network and init.d/functions are not very well adapted for Ash. Here are the two problems I found:

1) Use killproc to kill the given service process
   Steps:
```
	#!/bin/bash
	. /etc/init.d/functions

	test()
	{
		sleep 30 &
		killproc sleep
	}
	test
```
   run this testcase returns an error, and sleep is not killed:
	# ./test_killproc.sh: line 126: remainning+=2039 : not found

   This is because Ash cannot use "+=" for string concatenation.

2) I found out that here string(<<<) and c-style for-loop( for (( ; ; )) ) are
   used in network-scripts. But Ash doesn't support them either. So when I run:
   "/etc/rc.d/init.d/network status" gives these errors:
```
	# /etc/rc.d/init.d/network: ./network-functions: line 166: syntax error: unexpected redirection
	# /etc/rc.d/init.d/network: ./network-functions: line 575: syntax error: bad for loop variable
```
   The last two patches are used to fix this problem.

The problems mentioned above can all be verified simply on busybox v1.32.1. For example:
```
   $ /bin/busybox bash -c "str+=$char"
   bash: str+=: not found

   $ /bin/busybox bash -c "grep hello <<< "hello world""
   world: line 1: syntax error: unexpected redirection

   $ /bin/busybox bash -c "for (( idx=0; idx<5; idx++ )); do date; done"
   bash: syntax error: bad for loop variable
```

Thanks in advance.

### DIFF
--- a/etc/rc.d/init.d/functions
+++ b/etc/rc.d/init.d/functions
@@ -123,7 +123,7 @@ __kill_pids_term_kill_checkpids() {
         stime=${stat[21]}
 
         [ -n "$stime" ] && [ "$base_stime" -lt "$stime" ] && continue
-        remaining+="$pid "
+        remaining="$remaining$pid "
     done
 
     echo "$remaining"
--- a/network-scripts/ifup-eth
+++ b/network-scripts/ifup-eth
@@ -100,11 +100,11 @@ if [ "${TYPE}" = "Bridge" ]; then
           forward_delay="$(convert2sec ${forward_delay} centi)"
         fi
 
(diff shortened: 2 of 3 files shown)


jamacku (approved):
Great, Looks good to me! Thanks.

PR #403, opened 2022-01-12
ifup-routes: Log when using `ip $type replace`

jamacku:
Follow-up to #401


msekletar (approved):
LGTM

jamacku:
Thanks @msekletar!

PR #426, opened 2022-02-22
spec: Move `rename_device` to subpackage `initscripts-rename-device`

jamacku:
It allows to `rename_device` to be installed independently on `initscripts`.


msekletar (approved):
LGTM

PR #471, opened 2024-01-23
Fix splitting spaces in paths from statetab files

deajan:
Noticed that paths with spaces in `/etc/statetab` and `/etc/statetab.d/*` weren't interpreted.
Since a for loop was used, bash just splitted paths at default IFS chars.

Since we could have paths with spaces like `/opt/my service` in statetab, we need to read the statetab files line per line, with IFS being '\n', which can be accomplished by using a while loop.
This is already done for rwtab.

### DIFF
--- a/usr/libexec/readonly-root
+++ b/usr/libexec/readonly-root
@@ -184,17 +184,17 @@ if is_true "$READONLY" || is_true "$TEMPORARY_STATE"; then
 				mount -n --bind $bindmountopts "$STATE_MOUNT/$file" "$file"
 			fi
 
-			for path in $(grep -v "^#" "$file" 2>/dev/null); do
+			while read path ; do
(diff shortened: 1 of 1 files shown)


deajan:
PR initially fixed defect SC2013.
Got complaints from shellcheck about new SC2162 defect.

Initially wanted to code comply with current implementation of readonly-root script where rwtab's while loop didn't mangle backspaces (no SC2162 fix).

So updated this PR to fix SC2162 including the one from rwtab's while loop.

Has been tested on a couple of RHEL 9 setups with both rwtab and statetab entries.

jamacku (approved):
@deajan Thank you for the bug fix; it's very much appreciated.

I'm going to include it in the upcoming RHEL9 release.

PR #475, opened 2024-02-07
ifup-routes: print warning when ADDRESS0 entry is missing

suresh2514:
When static routes are added via route-<interface> file, the ifup-routes script will skip adding routes if entries not starting with ADDRESS0.

Good to print a warning so that users are aware of this.

### DIFF
--- a/network-scripts/ifup-routes
+++ b/network-scripts/ifup-routes
@@ -14,6 +14,13 @@ MATCH='^[[:space:]]*(\#.*)?$'
 handle_file () {
     . $1
     routenum=0
+
+    # print a warning if ADDRESS is not starting at ADDRESS0
+    if [ "x$(eval echo '$'ADDRESS$routenum)x" == "xx" ]; then
+        net_log $"Route entries are not sequentially ordered. Skipping $1 file" warning
+        exit 1
+    fi
+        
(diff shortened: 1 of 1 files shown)


jamacku (approved):
LGTM, thanks

network-scripts/ifup-routes
```diff
@@ -14,6 +14,13 @@ MATCH='^[[:space:]]*(\#.*)?$'
 handle_file () {
     . $1
     routenum=0
+
+    # print a warning if ADDRESS is not starting at ADDRESS0
+    if [ "x$(eval echo '$'ADDRESS$routenum)x" == "xx" ]; then
```
lnykryn, line 19: Looks, fine but two small nitpicks for the future. Since the routenum is always 0, the whole eval shenanigan is unnecessary. Simple [ -z "$ADDRESS" ]  would be much nicer.

network-scripts/ifup-routes
```diff
@@ -14,6 +14,13 @@ MATCH='^[[:space:]]*(\#.*)?$'
...
+        net_log $"Route entries are not sequentially ordered. Skipping $1 file" warning
+        exit 1
```
lnykryn, line 21: THis can potentially change the behaviour.

suresh2514:
Hi all,  

Thanks for the review and for accepting the patch.

regards

PR #480, opened 2024-04-16
fix: return support for network-scripts for Fedora 40

jamacku:
This commit partially reverts commit b751727763197b6d54735622274f644242141326 and e35db934845b01173280c1d05a9aa1d294f2627c

Resolves: [#2274830](https://bugzilla.redhat.com/show_bug.cgi?id=2274830)

### DIFF
--- a/initscripts.spec
+++ b/initscripts.spec
@@ -55,7 +55,6 @@ BuildRequires:    make
 BuildRequires:    systemd
 
 Obsoletes:        %{name}            < 10.16-1
-Obsoletes:        network-scripts    < 10.21-1
 
 # === PATCHES =================================================================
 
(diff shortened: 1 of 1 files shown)


initscripts.spec
```diff
@@ -116,6 +115,57 @@ This package provides service command.
...
+Requires(post):   chkconfig
+Requires(preun):  chkconfig
+
+Requires(post):   %{_sbindir}/update-alternatives
+Requires(postun): %{_sbindir}/update-alternatives
+
+Obsoletes:        %{name}            < 9.82-2
+
+# This is legacy and deprecated, so nobody should depend on this!
+# If ifcfg-style configuration is still desired, NetworkManager can do this.
```
lnykryn, line 146: Are we sure about this? I know that NM at least planned to drop support for ifcfg files

lnykryn (approved):

initscripts.spec
AdamWill: right now NetworkManager-initscripts-ifcfg-rh still exists in all current branches. I think there is an intent to retire it eventually, though. It is not installed by default since https://fedoraproject.org/wiki/Changes/NoIfcfgFiles (F36).

AdamWill:
I don't see any "if < 41" conditionals or anything in this, so doesn't this also return it to Rawhide? Is that intended?

jamacku:
@AdamWill I have returned `network-scripts` also to Rawhide, and I'm working on the Change Proposal.

PR #481, opened 2024-04-17
Fix build when %_bindir==%_sbindir

keszybz:
Preparation for https://fedoraproject.org/wiki/Changes/Unify_bin_and_sbin. The glob in %files wase causing a warning about duplicate entries in %files.


jamacku (approved):
LGTM, Thank you @keszybz

PR #485, opened 2024-08-27
Netconsole random fixes

lnykryn:


usr/libexec/netconsole
```diff
@@ -22,8 +22,7 @@ PATH=/sbin:/usr/sbin:$PATH
 RETVAL=0
 SERVER_ADDRESS_RESOLUTION=
 
-# Check that networking is up.
```
jamacku, line 25: I wouldn't remove the comment, but I don't have a strong opinion about it.

lnykryn: That comment was always misleading.

jamacku (approved):
LGTM, Thanks for the fix.

PR #490, opened 2024-09-25
retry arping one more time in ifup-eth when sendto failed

xanoxes:
retry arping one more time in ifup-eth when sendto failed

### DIFF
--- a/network-scripts/ifup-eth
+++ b/network-scripts/ifup-eth
@@ -293,10 +293,17 @@ else
 
             if ! LC_ALL=C ip addr ls ${REALDEVICE} | LC_ALL=C grep -q "${ipaddr[$idx]}/${prefix[$idx]}" ; then
                 if [ "${REALDEVICE}" != "lo" ] && ! is_false "${arpcheck[$idx]}"; then
-                    ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
-                    if [ $? = 1 ]; then
+                    for (( tries=0; tries<${ARPING_TRIES:=1}; tries++ )); do
+                        ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
+                        [ "$?" -eq 0 ] && break
                         ARPINGMAC=$(echo $ARPING |  sed -ne 's/.*\[\(.*\)\].*/\1/p')
-                        net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
+                        if [ -n "${ARPINGMAC}" ]; then
+                            net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
+                            exit 1
+                        fi
+                    done
+                    if [ "${tries}" -eq "${ARPING_TRIES}" ]; then
+                        net_log $"arping failed after $tries tries"
                         exit 1
                     fi
                 fi


xanoxes:
You're right, its ugly. So i refactor it with ARPING_TRIES. Please review this PR again.

network-scripts/ifup-eth
```diff
@@ -293,12 +293,22 @@ else
 
             if ! LC_ALL=C ip addr ls ${REALDEVICE} | LC_ALL=C grep -q "${ipaddr[$idx]}/${prefix[$idx]}" ; then
                 if [ "${REALDEVICE}" != "lo" ] && ! is_false "${arpcheck[$idx]}"; then
-                    ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
-                    if [ $? = 1 ]; then
-                        ARPINGMAC=$(echo $ARPING |  sed -ne 's/.*\[\(.*\)\].*/\1/p')
-                        net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
-                        exit 1
-                    fi
+                    while [ $tries -le $ARPING_TRIES ]
```
lnykryn, line 296: We tend to keep "while ...; do" on one line. Also, it is a good practice to always put variables in the statement into double quotes

network-scripts/ifup-eth
```diff
@@ -293,12 +293,22 @@ else
...
+                    do
+                        ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
+                        if [ $? = 1 ]; then
```
lnykryn, line 299: This will cause that "Error, some other host ".... message to be needlessly printed several times. It would be better to structure the code something like this:
while .. do
    arping
    ret=$?
done
if [ ret =1 ] ....

network-scripts/ifup-eth
```diff
@@ -293,12 +293,22 @@ else
...
+                            ARPINGMAC=$(echo $ARPING |  sed -ne 's/.*\[\(.*\)\].*/\1/p')
+                            net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
+                            tries=$((tries+1))
+                            if [ $tries -gt $ARPING_TRIES ]
```
lnykryn, line 303: Same as above, this should be checked after the cycle. And again missing double quotes

network-scripts/ifup-eth
```diff
@@ -293,12 +293,22 @@ else
...
+                            then
+                                net_log "arping failed after $tries tries"
```
lnykryn, line 305: Missing $ so the string could be translated  ($"string")

lnykryn (changes requested):

lnykryn:
And please squash your commits.

xanoxes:
@lnykryn

xanoxes:
@lnykryn

lnykryn:
> This will cause that "Error, some other host ".... message to be needlessly printed several times. It would be better to structure the code something like this: while .. do arping ret=$? done if [ ret =1 ] ....

I don't think this was addressed.

xanoxes:
> > This will cause that "Error, some other host ".... message to be needlessly printed several times. It would be better to structure the code something like this: while .. do arping ret=$? done if [ ret =1 ] ....
> 
> I don't think this was addressed.

We now only print "Error, some other host ".... once there really is an $ARPINGMAC in output of ARPING. In this case we exit the loop directly so there would not be a repeated Error msg print.

lnykryn:
> > > This will cause that "Error, some other host ".... message to be needlessly printed several times. It would be better to structure the code something like this: while .. do arping ret=$? done if [ ret =1 ] ....
> > 
> > 
> > I don't think this was addressed.
> 
> We now only print "Error, some other host ".... once there really is an $ARPINGMAC in output of ARPING. In this case we exit the loop directly so there would not be a repeated Error msg print.

Ah, I missed the break, sorry.

network-scripts/ifup-eth
```diff
@@ -293,10 +293,21 @@ else
...
+                    while [ "${tries}" -le "${ARPING_TRIES}" ]; do
+                        ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
+                        if [ $? = 1 ]; then
+                            ARPINGMAC=$(echo $ARPING |  sed -ne 's/.*\[\(.*\)\].*/\1/p')
+                            if [ -n "${ARPINGMAC}" ]; then
+                                net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
+                                break
+                            fi
+                            tries=$((tries+1))
+                        else
```
lnykryn, line 305: Can you flip this a bit? I think this would be nicer:
ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
[ $? = 0 ] && break
 ARPINGMAC=$(echo $ARPING |  sed -ne 's/.*\[\(.*\)\].*/\1/p')
....

network-scripts/ifup-eth
```diff
@@ -293,10 +293,18 @@ else
...
-                    ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
-                    if [ $? = 1 ]; then
+                    while [ "${tries}" -le "${ARPING_TRIES}" ]; do
+                        ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
+                        [ $? = 0 ] && break
                         ARPINGMAC=$(echo $ARPING |  sed -ne 's/.*\[\(.*\)\].*/\1/p')
-                        net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
+                        if [ -n "${ARPINGMAC}" ]; then
+                            net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
+                            break
```
lnykryn, line 302: Shouldn't it be here "exit 1" instead of "break"? If we find a duplicate address on the first iteration loop, then the if [ "${tries}" -gt "${ARPING_TRIES}" ]; check below will be false, and we will continue with the rest of the code.

lnykryn (changes requested):

xanoxes:
@lnykryn All done, plz check.

lnykryn:
I still think that there should be exit 1 instead of break on line 302 in your code

xanoxes:
> I still think that there should be exit 1 instead of break on line 302 in your code

YES, its wrong to use break, fixed.

xanoxes:
@lnykryn ALL fixed, plz check

network-scripts/ifup-eth
```diff
@@ -293,10 +293,18 @@ else
...
+                    while [ "${tries}" -le "${ARPING_TRIES}" ]; do
+                        ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
+                        [ $? = 0 ] && break
                         ARPINGMAC=$(echo $ARPING |  sed -ne 's/.*\[\(.*\)\].*/\1/p')
-                        net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
+                        if [ -n "${ARPINGMAC}" ]; then
+                            net_log $"Error, some other host ($ARPINGMAC) already uses address ${ipaddr[$idx]}."
+                            exit 1
+                        fi
+                        tries=$((tries+1))
```
jamacku, line 304: It will be handled in the head of the  `for` loop.

```suggestion
```

network-scripts/ifup-eth
```diff
@@ -293,10 +293,18 @@ else
...
+                    done
+                    if [ "${tries}" -gt "${ARPING_TRIES}" ]; then
```
jamacku, line 306: When using `for` loop, you need to update this condition

```suggestion
                    if [ "${tries}" -eq "${ARPING_TRIES}" ]  && [ "${arping_exit_code}" -ne 0 ]; then
```

network-scripts/ifup-eth
```diff
@@ -293,10 +293,18 @@ else
 
             if ! LC_ALL=C ip addr ls ${REALDEVICE} | LC_ALL=C grep -q "${ipaddr[$idx]}/${prefix[$idx]}" ; then
                 if [ "${REALDEVICE}" != "lo" ] && ! is_false "${arpcheck[$idx]}"; then
-                    ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
-                    if [ $? = 1 ]; then
+                    while [ "${tries}" -le "${ARPING_TRIES}" ]; do
+                        ARPING=$(/sbin/arping -c 2 -w ${ARPING_WAIT:-3} -D -I ${REALDEVICE} ${ipaddr[$idx]})
+                        [ $? = 0 ] && break
```
jamacku, line 298:
```suggestion
                        [ "${arping_exit_code:=$?}" -eq 0 ] && break
```

jamacku (changes requested):

xanoxes (commented):
LGTM

xanoxes:
@lnykryn All Fixed, plz check if any other issues.

xanoxes:
@lnykryn plz check

lnykryn (approved):

Issue #466, opened 2023-12-19
Busybox/bash does not support string subscripting, which causes an error.


AlfredWJ:
In some scenarios, scripts are executed using busybox/bash in the system environment, which does not support string subscripting. When encountering string subscripting, an error will be reported.
For example,the following code will result in an error message 'syntax error: bad substitution' when executed.
```
    read -r line < "/proc/self/stat"
    stat=$line
    stime=${stat[21]}
```
Can we modify it to the following?
```
diff --git a/etc/rc.d/init.d/functions b/etc/rc.d/init.d/functions
index 0d973b3..b4898be 100644
--- a/etc/rc.d/init.d/functions
+++ b/etc/rc.d/init.d/functions
@@ -119,8 +119,7 @@ __kill_pids_term_kill_checkpids() {
         [ ! -e  "/proc/$pid" ] && continue
         read -r line < "/proc/$pid/stat" 2> /dev/null

-        stat=($line)
-        stime=${stat[21]}
+        stime=$(cat /proc/$pid/stat | cut -d" " -f22)

         [ -n "$stime" ] && [ "$base_stime" -lt "$stime" ] && continue
         remaining="$remaining$pid "
@@ -141,8 +140,7 @@ __kill_pids_term_kill() {

     # We can't initialize stat & base_stime on the same line where 'local'
     # keyword is, otherwise the sourcing of this file will fail for ksh...
-    stat=($(< /proc/self/stat))
-    base_stime=${stat[21]}
+    base_stime=$(cat /proc/self/stat | cut -d" " -f22)

     if [ "$1" = "-d" ]; then
         delay=$2
```

lnykryn:
```
-    stat=($(< /proc/self/stat))
-    base_stime=${stat[21]}
+    base_stime=$(cat /proc/self/stat | cut -d" " -f22)
```

This is not equivalent, old code returns stime for the script while the new one returns the time for the "cat".

AlfredWJ:
> ```
> -    stat=($(< /proc/self/stat))
> -    base_stime=${stat[21]}
> +    base_stime=$(cat /proc/self/stat | cut -d" " -f22)
> ```
> 
> This is not equivalent, old code returns stime for the script while the new one returns the time for the "cat". You probably want cat /proc/$$/stat.
I try cat /proc/$$/stat, and it is not equivalent too. Can we make the modification like this:
```
-    stat=($(< /proc/self/stat))
-    base_stime=${stat[21]}
[...]

PR #498, opened 2025-06-25
functions: avoid using array indices to get value form string

AlfredWJ:
Busybox-bash does not support "[]", fix it by "cut"

issue: https://github.com/fedora-sysv/initscripts/issues/466


AlfredWJ:
image
is it stuck？

